PR #2, opened 2025-03-28
Area of intersection operator implementation

asinghvi17:
This requires `]add GeometryOps#as/trees-and-antimeridian GeometryOpsCore#as/trees-and-antimeridian`.  

We could move the dual tree query code here and let this work on the latest released version of GeometryOps, but that code is a bit tough.

### DIFF
--- a/src/ConservativeRegridding.jl
+++ b/src/ConservativeRegridding.jl
@@ -2,6 +2,9 @@ module ConservativeRegridding
 
 using DocStringExtensions
 
+import GeometryOps as GO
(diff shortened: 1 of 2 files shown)


milankl:
You're too quick, I already copied your code into mk/init and now we've kind of done similar things but you've added some more stuff. Can I merge #1 first and then we do things bit by bit?

asinghvi17:
Yeah that sounds better.  I'll remove the changes from regridder.jl here and keep the rest.

PR #3, opened 2025-03-28
add compat with antimeridian branch

milankl:
Is this right @asinghvi17 ?


asinghvi17:
Don't need the URL (I think)

milankl:
Interesting, Pkg rewrote this automatically to
```julia
[sources.GeometryOps]
rev = "as/trees-and-antimeridian"

[sources.GeometryOpsCore]
rev = "as/trees-and-antimeridian"
```

PR #17, opened 2025-04-15
Regrid on domains with different extent

simone-silvestri:
by using the intersection areas to infer the areas of the polygons, if the two domains have different extent that does not overlap completely, some areas will be zero leading to NaNs where there is no intersection. 

This PR computes the areas directly from the polygons to avoid this issue


simone-silvestri:
I think we can merge this for the moment.

PR #19, opened 2025-05-01
Update GeometryOps compat

asinghvi17:
This is now ready for release, although antimeridian cutting will need more work.  Still that's somewhat outside the scope of this package.


glwagner:
Is the anti-meridian cutting the last thing we need to get to a first prototype? Which package is it in scope for?

asinghvi17:
Yeah, that's the last thing.  That's more on the GeometryOps side.

asinghvi17:
Doh I already did that in here @juliasloan25

PR #29, opened 2025-12-05
Add a ClimaCore test

juliasloan25:
We'd like to use ConservativeRegridding.jl in ClimaCoupler.jl for coupled simulations. This requires some compatibility with ClimaCore spaces/fields. This PR adds a preliminary prototype/test to regrid between two ClimaCore fields.

Note that this test makes a big over-simplification that will need to be addressed. We take the nodal values of the spectral element ClimaCore field and average them to get one value per element, as we would have for finite volume. This allows us to use the current ConservativeRegridding infrastructure, but isn't what we want to do long term. We'll start with this and continue to iterate on it.

Co-authored with @ph-kev
cc @simone-silvestri

### DIFF
--- a/test/climacore.jl
+++ b/test/climacore.jl
@@ -176,7 +176,7 @@ set_value_per_element!(field1_one_value_per_element, value_per_element1)
 
 # Test our helper functions
 # Check that integrating over each element and summing gives the same result as integrating over the whole domain
-@test isapprox(sum(integrate_each_element(field1)), sum(field1), atol = 1e-12)
+@test isapprox(sum(integrate_each_element(field1)), sum(field1), atol = 1e-11)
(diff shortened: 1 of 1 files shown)


juliasloan25:
> This looks broadly reasonable but it would be good to understand your use case in more detail. I'd discussed this with @simone-silvestri last JuliaEO but not sure if things might have changed since then? Also, there are things we should figure out with respect to spherical/planar...
> 
> If it's easier we could just do a call, I'm in EST so it should be easy to coordinate. Most afternoons EST work for me.

I think a call would be helpful! Simone and I are planning to meet this upcoming Tuesday at 12pm EST. You're welcome to join then, or I could do Monday 1-3pm EST or later in the week

PR #35, opened 2026-01-07
Storage space to regrid non-contiguous memory

simone-silvestri:
if we try to use the regridder to regrid non contiguous data, the matrix multiply uses methods for full matrices rather than sparse matrices. 
To reproduce:
```julia
a = rand(210, 210)
b = rand(210, 210)

ad = Array(view(a, 1:200, 1:200))
bd = Array(view(b, 1:200, 1:200))
A = sprand(Float64, 200*200, 200*200, 0.0001)
```
then
```julia
julia> @time LinearAlgebra.mul!(vec(view(a, 1:200, 1:200)), A, vec(view(b, 1:200, 1:200)));
  7.000757 seconds (4 allocations: 320 bytes)

julia> @time LinearAlgebra.mul!(vec(ad), A, vec(bd));
  0.000567 seconds (2 allocations: 64 bytes)
```
This PR allocates a storage space in the regridder (like in XESMF.jl) to allow regridding non-contiguous memory efficiently.

### DIFF
--- a/src/regrid.jl
+++ b/src/regrid.jl
@@ -41,38 +41,10 @@ function regrid!(dst_field::DenseVector, regridder::Regridder, src_field::Abstra
 end
 
 function regrid!(dst_field::AbstractVector, regridder::Regridder, src_field::DenseVector)
-    regrid!(regridder.dst_tmp, regridder, src_field)
+    regrid!(regridder.dst_temp, regridder, src_field)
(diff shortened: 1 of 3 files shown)


asinghvi17:
Already merged into the other PR but I have to rethink the regridder construction interface a bit

asinghvi17:
Merged into #34

PR #46, opened 2026-01-19
Try to add a sweat test

asinghvi17:
The idea is to try regridding each grid against each other grid we support to see where problems crop up.  I think this found quite a few though


asinghvi17:
Wait what?!  It completely fails locally

PR #47, opened 2026-01-22
Implement `on_architecture` method for a Regridder

simone-silvestri:
required if we want to regrid on the GPU, not sure we can test this on GHA because they do not have GPUs


asinghvi17:
Seems good to me.  Can we test with `pocl_jll` and OpenCL?

milankl:
I still think we should put a `KernelLaunching` module into KernelAbstractions that implements all these methods around `architecture`, `on_architecture`, the work orders and `launch!`

PR #60, opened 2026-02-10
General improvements

asinghvi17:
- Better comments / docs and CLAUDE.md
- Remove Oceananigans branch
- Reactivate 2 out of 4 ClimaCore extension tests

### Rename `FaceAwareQuadtreeCursor` → `IndexOffsetQuadtreeCursor`
Store the raw index offset instead of a face index, removing the assumption that all faces have equal size. This makes the cursor usable for any multi-grid scenario with arbitrary offsets.

### Add `MultiTreeWrapper` for composite spatial trees
A non-leaf spatial tree node that holds multiple sub-trees as children. Each sub-tree handles its own global-to-local index conversion (e.g., via `IndexOffsetQuadtreeCursor`). Not an `AbstractTreeWrapper` subtype since it doesn't wrap a single tree — this lets the generic `areas()` fallback work without a special override.

### Add fold-aware `treeify` for `TripolarGrid` with `RightFaceFolded`
Split the grid into three sub-trees: the main body (rows 1:Ny-1), and the left/right halves of the fold row, each wrapped in an `IndexOffsetQuadtreeCursor` inside a `MultiTreeWrapper`. This separates the fold topology from the regular grid for cleaner spatial indexing.

### DIFF
--- a/ext/ConservativeRegriddingOceananigansExt.jl
+++ b/ext/ConservativeRegriddingOceananigansExt.jl
@@ -165,31 +165,11 @@ function Trees.treeify(
     #    ghost cells via PaddedTreeWrapper.  Ghost cells exist for dimension
     #    matching (the source field has Nhalf cells per fold half) but are never
     #    reached during the dual DFS, so they contribute nothing to intersections.
-    #    We use ExplicitPolygonGrid (not CellBasedGrid) so that each cell's
-    #    polygon is independent.
     Nquarter = Nhalf ÷ 2
 
-    # Build ExplicitPolygonGrids with only the unique cells (Nquarter per half)
-    left_polys = Matrix{GI.Polygon}(undef, Nquarter, 1)
-    for k in 1:Nquarter
-        left_polys[k, 1] = GI.Polygon(SA[GI.LinearRing(SA[
-            cells_unitspherical[k, Ny], cells_unitspherical[k+1, Ny],
-            cells_unitspherical[k+1, Ny+1], cells_unitspherical[k, Ny+1],
-            cells_unitspherical[k, Ny]
-        ])])
-    end
-    left_real_grid = Trees.ExplicitPolygonGrid(manifold, left_polys)
-
-    right_polys = Matrix{GI.Polygon}(undef, Nquarter, 1)
-    for k in 1:Nquarter
-        i = Nhalf + k
-        right_polys[k, 1] = GI.Polygon(SA[GI.LinearRing(SA[
-            cells_unitspherical[i, Ny], cells_unitspherical[i+1, Ny],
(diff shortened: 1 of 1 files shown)


ext/ConservativeRegriddingOceananigansExt.jl
```diff
@@ -59,6 +85,48 @@ end
 
 
 # Define the ConservativeRegridding interface for Oceananigans grids.
+
+function Trees.treeify(
+    manifold::GOCore.Spherical,
+    grid::Oceananigans.Grids.ZRegOrthogonalSphericalShellGrid{<: Number, <: Any, Oceananigans.RightCenterFolded}
```
asinghvi17:
```suggestion
    grid::Oceananigans.Grids.ZRegOrthogonalSphericalShellGrid
```

Just to do it for all grids.  We should try with both kinds and do a regridding though.

test/extensions/climacore.jl
```diff
@@ -12,24 +12,24 @@ const ClimaCoreExt = Base.get_extension(ConservativeRegridding, :ConservativeReg
...
         n_quad_points = 2,
         h_elem = 64,
     )
 
     # Define a field on the first space, to use as our source field
-    field = Fields.coordinate_field(cubedsphere_space).long
-    ones_field = Fields.ones(cubedsphere_space)
+    field = Fields.coordinate_field(cubedsphere_space).long .+ 180
+    ones_field = ones(cubedsphere_space)
     cubed_sphere_vals = zeros(6*cubedsphere_space.grid.topology.mesh.ne^2)
```
juliasloan25:
```suggestion
    cubed_sphere_vals = zeros(Meshes.nelements(cubedsphere_space.grid.topology.mesh))
```

ext/ConservativeRegriddingOceananigansExt.jl
```diff
@@ -58,7 +82,155 @@ function _compute_cell_matrix!(cell_matrix, Nx, Ny, ℓx, ℓy, grid)
...
+    right_polys = Matrix{GI.Polygon}(undef, Nquarter, 1)
+    for k in 1:Nquarter
+        i = Nhalf + k
+        right_polys[k, 1] = GI.Polygon(SA[GI.LinearRing(SA[
+            cells_unitspherical[i, Ny], cells_unitspherical[i+1, Ny],
+            cells_unitspherical[i+1, Ny+1], cells_unitspherical[i, Ny+1],
+            cells_unitspherical[i, Ny]
+        ])])
+    end
+    right_real_grid = Trees.ExplicitPolygonGrid(manifold, right_polys)
```
asinghvi17: Just need to clean this up - it should not be doing this.

PR #69, opened 2026-03-01
Fix planar grid threading and mapreduce argument order

### LINKED ISSUE #65: mapreduce arguments swapped in cell_range_extent for ExplicitPolygonGrid{Planar}
## Description

In `src/trees/grids.jl:130`, the `mapreduce` call in `cell_range_extent` for `ExplicitPolygonGrid{<: GO.Planar}` has the `f` and `op` arguments swapped:

```julia
function cell_range_extent(q::ExplicitPolygonGrid{<: GO.Planar}, irange::UnitRange{Int}, jrange::UnitRange{Int})
    return mapreduce(Extents.union, GI.extent, (getcell(q, i, j) for i in irange, j in jrange))
end
```

Julia's `mapreduce(f, op, itr)` expects `f` as the mapping function and `op` as the binary reducing operation. The current code passes `Extents.union` (a binary op) as the mapping function and `GI.extent` (a unary function) as the reducing operation.

**Fix**: Swap the arguments:
```julia
return mapreduce(GI.extent, Extents.union, (getcell(q, i, j) for i in irange, j in jrange))
```

Note: the analogous `Spherical` method on line 133 does not use `mapreduce` and is unaffected.

## MWE

```julia
using ConservativeRegridding
using GeoInterface
using GeometryOpsCore

function make_grid(nx, ny)
    polys = Matrix{GeoInterface.Polygon}(undef, nx, ny)
    for j in 1:ny, i in 1:nx
        x0, x1 = (i-1)/nx, i/nx
        y0, y1 = (j-1)/ny, j/ny
        ring = GeoInterface.LinearRing([(x0,y0),(x1,y0),(x1,y1),(x0,y1),(x0,y0)])
        polys[i,j] = GeoInterface.Polygon([ring])
    end
    polys
end

src_grid = make_grid(2, 2)
dst_grid = make_grid(3, 3)

# This errors with: MethodError: no method matching union(::GeoInterface.Wrappers.Polygon{...})
r = ConservativeRegridding.Regridder(GeometryOpsCore.Planar(), dst_grid, src_grid; threaded=false)
```

## Error

```
MethodError: no method matching union(::GeoInterface.Wrappers.Polygon{...})
```

## Version

ConservativeRegridding v0.2.0

### LINKED ISSUE #66: _area_criterion not implemented for Planar extents (threaded=true fails)
## Description

In `src/regridder/intersection_areas.jl:35`, the `_area_criterion` function for `Extents.Extent` (used with Planar grids) explicitly throws an error:

```julia
_area_criterion(cap::Extents.Extent) = error("Area criterion for multithreading has to be customizable, for 2D planar extents - this needs to be implemented!")
```

This means using `threaded=true` (the default) with Planar grids will always fail once the multithreaded dual DFS calls `_area_criterion` on the tree node extents.

The `Spherical` path has a working implementation for `SphericalCap` on line 34.

Note: reproducing this in isolation requires first fixing #65 (the `mapreduce` argument swap), since tree construction fails before reaching the threading code path.

## MWE

```julia
using ConservativeRegridding
using GeoInterface
using GeometryOpsCore
using Extents

function make_grid(nx, ny)
    polys = Matrix{GeoInterface.Polygon}(undef, nx, ny)
    for j in 1:ny, i in 1:nx
        x0, x1 = (i-1)/nx, i/nx
        y0, y1 = (j-1)/ny, j/ny
        ring = GeoInterface.LinearRing([(x0,y0),(x1,y0),(x1,y1),(x0,y1),(x0,y0)])
        polys[i,j] = GeoInterface.Polygon([ring])
    end
    polys
end

src_grid = make_grid(2, 2)
dst_grid = make_grid(3, 3)

# First, fix bug #65 so we can reach the threading code path:
@eval function ConservativeRegridding.Trees.cell_range_extent(
    q::ConservativeRegridding.Trees.ExplicitPolygonGrid{<:GeometryOpsCore.Planar},
    irange::UnitRange{Int}, jrange::UnitRange{Int},
)
    return mapreduce(
        GeoInterface.extent, Extents.union,
        (ConservativeRegridding.Trees.getcell(q, i, j) for i in irange, j in jrange),
    )
end

# This errors with: "Area criterion for multithreading has to be customizable..."
r = ConservativeRegridding.Regridder(GeometryOpsCore.Planar(), dst_grid, src_grid; threaded=true)
```

## Error

```
ErrorException: Area criterion for multithreading has to be customizable, for 2D planar extents - this needs to be implemented!
```

## Workaround

Use `threaded=false`:
```julia
r = ConservativeRegridding.Regridder(GeometryOpsCore.Planar(), dst_grid, src_grid; threaded=false)
```

## Version

ConservativeRegridding v0.2.0

ctessum-claude:
## Summary

Fixes two bugs that prevent `Planar` grid regridding from working with `threaded=true` (the default):

- **Fix #65**: In `src/trees/grids.jl:130`, `mapreduce(Extents.union, GI.extent, ...)` had swapped arguments. The correct call is `mapreduce(GI.extent, Extents.union, ...)` where `GI.extent` is the mapping function and `Extents.union` is the reduction operator.
- **Fix #66**: In `src/regridder/intersection_areas.jl:35`, `_area_criterion(cap::Extents.Extent)` threw an unimplemented error. Now returns `true` to always allow task spawning for planar extents, since the tree structure itself handles granularity control.

## Test plan

- [x] Added regression test that creates planar polygon grids and constructs a `Regridder` with `threaded=true`, verifying it succeeds and produces valid output
- [ ] Full test suite passes


asinghvi17:
Hmm, this wouldn't really do anything...

https://github.com/JuliaGeo/ConservativeRegridding.jl/blob/dcba88685f5e4cd76c3af574e002565ac896d529/src/utils/MultithreadedDualDepthFirstSearch.jl#L15-L17

this means that at the first invocation of multithreaded dfs on the top level trees, it would immediately go to single threaded with the whole tree, not even performing an initial split.  I don't think we should change that in the model.

I'm not sure what the best API is here but somehow we'd have to have a tree specific way of passing down some area criterion.  Perhaps a wrapper would work?  And there could be a `get_area_criterion` method that gets invoked on `(manifold, tree)`, which provides some object that will perform the query.  What do you think @ctessum?

PR #71, opened 2026-03-05
Fix loop order in _compute_cell_matrix! function

navidcy:
This is based on https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-column-major


asinghvi17 (approved):
Seems reasonable and should not be changing any values, will wait for @simone-silvestri though and then merge!

simone-silvestri (approved):

PR #77, opened 2026-03-14
Add XESMF.jl comparison test

asinghvi17:
## Summary
- Adds a test comparing ConservativeRegridding against XESMF.jl (Python xESMF/ESMF) cell-by-cell on regular lon/lat grids
- Tests three configurations: coarse→fine, fine→coarse, and same-resolution using Oceananigans `LatitudeLongitudeGrid`
- Compares both the sparse weight matrices (after normalizing CR's raw areas by destination cell area) and the regridded output for a smooth test field
- Both regridders agree to machine precision (~1e-14 relative error on weights, ~1e-15 on fields)

## Test plan
- [x] `julia --project=test -e 'include("test/usecases/xesmf_comparison.jl")'` passes (12/12 tests)
- [x] Full test suite passes with the new test added to `runtests.jl`


asinghvi17:
Ok can we initialize the Oceananigans grids with a radius of 1 such that this passes to xesmf and also the manifold in the conservativeregridding call

PR #79, opened 2026-03-25
Unify Tripolar grid extension

simone-silvestri:
after https://github.com/CliMA/Oceananigans.jl/pull/5408, the `RightFaceFolded` tripolar grid went back to having a prognostic domain all the way up to `1:Ny` (it was previously `1:Ny-1`).

Therefore, all the shenanigans for a `RightFaceFolded` tripolar grid are not required anymore. Now the difficult one is the `RightCenterFolded` which has an extra half row at `j = Ny`. The `RightFaceFolded` should be the easy one.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -41,7 +41,7 @@ ConservativeRegriddingSpeedyWeatherExt = ["RingGrids", "SpeedyWeather"]
 [compat]
 ChunkSplitters = "3"
 ClimaCore = "0.14.49"
-Oceananigans = "0.104, 0.105"
+Oceananigans = "0.106"
 DocStringExtensions = "0.8, 0.9"
 Extents = "0.1"
 GeoInterface = "1"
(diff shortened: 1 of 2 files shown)


asinghvi17:
looks like this passes the sweat tests!  But tripolar is a bit of a special case there since you can't test integral conservation, so I will try to add some more.  Perhaps simply testing conservation in the northern hemisphere is sufficient somehow.

simone-silvestri:
@asinghvi17 you want more tests or should we merge this?

juliasloan25 (approved):
Since `RightCenterFolded` is the more complicated option now, do we need to add in some infrastructure to handle it? Or error for now if it isn't supported?

simone-silvestri:
@asinghvi17 I think we can merge this and look at more comprehensive tests in another PR

asinghvi17:
Sounds good.  I am running the sweat test pr on top of these changes now, so we should see any major regression there.

PR #82, opened 2026-03-27
implement correct SE regridding

### LINKED ISSUE #81: Improve ClimaCore spectral element regridding
Right now when we regrid to/from a (spectral element) ClimaCore space, we approximate it as finite volume. That is, we average the value over an element and take that to be the value over the entire element. This is a very poor approximation, and now that we have all the software interfaces working we need to improve it.

The full math is written up in [this document](https://www.overleaf.com/read/wxdjqcgcnqqp#99a680) - here we just use the results and begin converting it into code.

Note: the approaches described here use the "node-in-polygon" approach described in that document. A major drawback to this approach is that if source and target grids have very different resolutions, we may have no source node within the destination element. This would produce 0 for those target nodes/cells - we should include a backup for this case.

## source SE -> target FV
For a FV cell `k`, we want to compute:
Image

We choose weights `B` as
Image
where `k` is a FV cell, `e` is is a SE element, `i, j` are the nodal indices within element `e`, `W_{ij}^e` is the integration weight of element `e` at `i,j`, and x_{ij}^e is the node in element `e` at indices `i,j`.

Substituting this back in we get
Image

That is, for each FV cell, we take a jacobian-weighted integral of the SE nodes that fall inside it, then normalize by the destination cell area.

### pseudocode
```
for each SE element e:
    for each FV cell k intersecting e:
        for each SE node (i,j) in e:
            if x[e,i,j] is inside k:
                f_dst[k] += W[e,i,j] * f_src[e,i,j]

for each FV cell k:
    f_dst[k] /= A[k]
```

## source FV -> target SE
This case is simpler - each SE node gets the value of the FV cell that contains it. We end up with
Image

### pseudocode
```
for each SE element e:
    for each FV cell k intersecting e:
        for each SE node (i,j) in e:
            if x[e,i,j] is inside k:
                f_dst[e,i,j] = f_src[k]
```

## source SE -> target SE
This case is analogous to the SE -> FV case
Image

### pseudocode
```
# Compute destination element areas
for each destination element e':
    A[e'] = sum of W[e',i',j'] over all nodes (i',j') in e'

# Accumulate source mass into destination elements
for each source element e:
    for each destination element e' intersecting e:
        for each source node (i,j) in e:
            if x[e,i,j] is inside e':
                numerator[e'] += W[e,i,j] * f_src[e,i,j]

# Normalize by destination element area
for each destination element e':
    for each destination node (i',j') in e':
        f_dst[e',i',j'] = numerator[e'] / A[e']
```

juliasloan25:
Closes #81

Implements spectral element regridding for ClimaCore spaces, rather than just approximating them as finite volume grids.

## To do
- [ ] make sure tests are strict enough
- [ ] check indexing
- [ ] fix SE -> FV regridding
- [ ] remove outdated use_cases


src/regridder/se_regridder.jl
```diff
@@ -0,0 +1,150 @@
+"""
+    SEtoFVRegridder{W,A,V} <: AbstractRegridder
+
+Regridder from a spectral element (SE) grid to a finite volume (FV) grid.
+
+The `weight_matrix` is a sparse matrix of size `(N_fv, N_se_nodes)` whose entries
+are the SE Jacobian integration weights `W[e,i,j]` for each SE node that falls
+inside a given FV cell.  Regridding computes `dst = weight_matrix * src / dst_areas`.
+"""
+struct SEtoFVRegridder{W, A, V} <: AbstractRegridder
```
juliasloan25, line 10: reuse default Regridder

docs/src/extensions/climacore.md
```diff
@@ -0,0 +1,147 @@
...
+
+# ClimaCore field → flat vector
+src_vec = Ext.se_field_to_vec(field)
+
+# Flat vector → ClimaCore field
+dst_field = Fields.zeros(space)
+Ext.vec_to_se_field!(dst_field, dst_vec)
+
+# Node positions and weights (for diagnostics / verification)
+positions = Ext.se_node_positions(space)   # Vector{UnitSphericalPoint}
```
asinghvi17:
```suggestion
using ClimaCore: Fields
const ClimaCoreExt = Base.get_extension(ConservativeRegridding, :ConservativeRegriddingClimaCoreExt)

# ClimaCore field → flat vector
src_vec = ClimaCoreExt.se_field_to_vec(field)

# Flat vector → ClimaCore field
dst_field = Fields.zeros(space)
ClimaCoreExt.vec_to_se_field!(dst_field, dst_vec)

# Node positions and weights (for diagnostics / verification)
positions = ClimaCoreExt.se_node_positions(space)   # Vector{UnitSphericalPoint}
```

asinghvi17:
The merge was mostly just pulling things in from the sweat test PR

asinghvi17:
@juliasloan25 are you still working on this one?

juliasloan25:
> @juliasloan25 are you still working on this one?

Yes - I was focused on some other things and then on vacation, but I'll return to this next week

src/regridder/se_regridder.jl
juliasloan25: We can't do this because `regrid!` takes in generic Array types. However the `Regridder` constructor is extended to construct the correct regridder type under the hood, so users don't need to know there are multiple types

juliasloan25:
superseded by #99

PR #83, opened 2026-04-12
Fix SphericalCap extent bugs for CellBasedGrid with UnitSphericalPoint vertices

cfranken:
## Summary

Two bugs in `circle_from_four_corners` (`src/trees/grids.jl`) that surface when `CellBasedGrid{Spherical}` stores `UnitSphericalPoint` vertices directly (instead of geographic `(lon, lat)` tuples):

1. **UnitSphericalPoint double-conversion:** `circle_from_four_corners` unconditionally ran `GO.UnitSphereFromGeographic()` on all input points, including points that were already `UnitSphericalPoint`s. This reinterpreted Cartesian `(x, y, z)` as geographic `(lon°, lat°)` — producing nonsensical coordinates that poisoned `SphericalCap` extents.

2. **Degenerate center vector → NaN:** When the four corners of a cell-range extent are symmetrically placed around the origin (e.g., a latitude band straddling the equator with exactly 180° longitude span), their Cartesian sum is `(0, 0, 0)` and `normalize((0,0,0))` produces `NaN`. This silently poisoned SphericalCap radii, causing the dual DFS to prune entire subtrees (NaN comparisons always return `false`). Also added a safe midpoint fallback for `slerp` on near-antipodal point pairs.

## Fix

- Introduce `_to_unit_sphere(pt)`: passes through `UnitSphericalPoint`s unchanged, only converts geographic tuples via `UnitSphereFromGeographic()`.
- Detect near-zero center norm (`< 1e-12`), fall back to centroid of all border points, and use an arbitrary axis `(0, 0, 1)` if that also cancels.
- Guard `slerp` against near-antipodal inputs (dot product < -0.985, i.e., angle > ~170°) with a normalized Euclidean midpoint.

## How it was found

While integrating CR.jl into [AtmosTransport.jl](https://github.com/RemoteSensingTools/AtmosTransport.git) for conservative regridding between LatLon / CubedSphere / ReducedGaussian meshes, we build per-ring `CellBasedGrid{Spherical}` trees with `UnitSphericalPoint` corner matrices. The equatorial ring (lat ±15°, 16 cells spanning 0°–360°) triggered both bugs simultaneously: the first 8-cell sub-range covers lon 0°–180° at lat ±15°, which has exact center cancellation.

The Oceananigans extension (`ConservativeRegriddingOceananigansExt.jl`) also stores `UnitSphericalPoint` vertices in `CellBasedGrid`, so this bug potentially affects any spherical grid where a cell-range extent happens to span >= 180° of longitude while straddling the equator.

## Testing

Tested via AtmosTransport's `test_v2/regridding/runtests.jl` (564 tests, all passing), which exercises LatLon ↔ CubedSphere ↔ ReducedGaussian conservative regridding with mass conservation checks to machine precision (rel err ~1e-16 for constant fields).

I'd suggest also adding a dedicated test to CR.jl's test suite for this case: build a `CellBasedGrid{Spherical}` with `UnitSphericalPoint` vertices on a lat band crossing the equator with >= 180° longitude span, verify `cell_range_extent` returns a finite non-NaN `SphericalCap`, and confirm that the `Regridder` finds nonzero intersections.

## Related issues and future PRs we'd be happy to contribute

While integrating CR.jl, we identified several additional areas where the package could be improved. We're happy to contribute PRs for any of these if useful:

1. **Missing `_intersects(Extent, SphericalCap)` method (in GeometryOps):** `FlatNoTree` over polygons with `UnitSphericalPoint` vertices produces `Extent{(:X,:Y,:Z)}` per-cell extents, but the dual DFS uses `_intersects(SphericalCap, SphericalCap)` as the predicate on spherical manifolds. The cross-type comparison crashes with `MethodError`. We worked around it by using `MultiTreeWrapper` + `CellBasedGrid` per ring instead of `FlatNoTree`, but the missing method is a footgun for any user who tries `FlatNoTree` on spherical polygons.

2. **Built-in ESMF offline-weights export:** `save_esmf_weights(path, regridder)` writing the standard `(S, row, col, frac_a, frac_b, area_a, area_b)` NetCDF format. We implemented this in ~70 lines in our wrapper; it could live in CR.jl's core for cross-tool validation against xESMF.

3. **Raw-corner cubed-sphere `treeify`:** A `treeify(Spherical(), corner_matrices::NTuple{6, Matrix{UnitSphericalPoint}})` method in the core package, so users can build a `CubedSphereToplevelTree` from 6 face-corner arrays without depending on ClimaCore.

4. **JLD2 save/load example in the docs:** The `Regridder` struct is trivially serializable, but this isn't documented anywhere.

5. **Benchmarks at realistic scale:** A table of (grid pair, nnz, build time, threads) in the docs would help users evaluating CR.jl for production grids (C180, C720, N320).

6. **`normalize=true` default + the TODO:** The default `normalize=true` divides by `max(intersections)`, making weight-level comparisons with xESMF impossible. The code has a `# TODO is this the best normalizer?` comment. At minimum the docstring should warn about this.

7. **`areas()` ordering contract:** The implicit contract that `areas[k]` matches sparse-matrix index `k` is undocumented and fragile for composite trees.

8. **`RegularGrid{Spherical}` with lon-lat tuples:** It's unclear whether `GO.area(Spherical(), polygon_with_lonlat_tuples)` is correct, since the Oceananigans extension always converts to `UnitSphericalPoint` first.

Co-authored-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/trees/grids.jl
+++ b/src/trees/grids.jl
@@ -172,18 +172,59 @@ of the given range, and then expand it such that it encloses all other points al
 Since we know our grids are curvilinear, this is broadly okay.  It's not perfectly efficient, though.
 =#
 
-using LinearAlgebra
+"""
+    _safe_slerp(a, b, t) -> UnitSphericalPoint
+
+Spherical linear interpolation with a fallback for near-antipodal points.
+
+When `dot(a, b) < -0.985` (angle > ~170°), standard `slerp` divides by
+`sin(angle) ≈ 0`, producing NaN or Inf. In that case, fall back to the
+normalized Euclidean midpoint. If that also degenerates (exactly antipodal),
+return `a` unchanged — the resulting cap radius will be slightly loose but
+finite and correct for pruning purposes.
+"""
+function _safe_slerp(a, b, t)
+    d = LinearAlgebra.dot(a, b)
+    if d > -0.985  # angle < ~170°, slerp is safe
+        return GO.UnitSpherical.slerp(a, b, t)
+    else
+        mid = (1 - t) .* a .+ t .* b
+        n = LinearAlgebra.norm(mid)
+        return n < 1e-14 ? a : LinearAlgebra.normalize(mid)
+    end
+end
+
 function circle_from_four_corners(corner_points, other_points)
     raw = GO.UnitSphereFromGeographic().(corner_points)
     # corner_points arrive as (BL, TL, BR, TR); reorder to CCW (BL, BR, TR, TL)
     # so that consecutive slerps give bottom, right, top, left edge midpoints.
     p1, p2, p3, p4 = raw[1], raw[3], raw[4], raw[2]
-    center = LinearAlgebra.normalize((p1 .+ p2 .+ p3 .+ p4) ./ 4)
+    raw_center = (p1 .+ p2 .+ p3 .+ p4) ./ 4
+    nrm = LinearAlgebra.norm(raw_center)
+    # When the four corners are symmetrically placed around the origin
+    # (e.g., a latitude band straddling the equator with exactly 180° of
+    # longitude span), their Cartesian sum cancels to near-zero and
+    # normalize() produces NaN. Fall back to the centroid of ALL border
+    # points (which break the symmetry because they include intermediate
+    # longitude samples), or an arbitrary axis if everything cancels.
+    #
+    # Threshold: 1e-12 is ~eps(Float64) × 1e4, catching sums that are
+    # zero up to floating-point cancellation of O(1) vectors while not
+    # triggering on legitimate small-but-nonzero centroids.
+    if nrm < 1e-12
+        other_raw = GO.UnitSphereFromGeographic().(other_points)
+        all_pts = vcat(collect(raw), collect(other_raw))
+        raw_center = sum(all_pts) ./ length(all_pts)
+        nrm = LinearAlgebra.norm(raw_center)
+    end
+    center = nrm < 1e-14 ?
+        GO.UnitSpherical.UnitSphericalPoint(0.0, 0.0, 1.0) :
+        LinearAlgebra.normalize(raw_center)
     # Midpoints of edges (bottom, right, top, left)
-    p12 = GO.UnitSpherical.slerp(p1, p2, 0.5)
-    p23 = GO.UnitSpherical.slerp(p2, p3, 0.5)
-    p34 = GO.UnitSpherical.slerp(p3, p4, 0.5)
-    p41 = GO.UnitSpherical.slerp(p4, p1, 0.5)
+    p12 = _safe_slerp(p1, p2, 0.5)
+    p23 = _safe_slerp(p2, p3, 0.5)
+    p34 = _safe_slerp(p3, p4, 0.5)
+    p41 = _safe_slerp(p4, p1, 0.5)
     # Distance to the furthest corner or edge midpoint
     corner_distance = maximum(p -> GO.spherical_distance(center, p), (p1, p2, p3, p4, p12, p23, p34, p41))
     # Distance to the furthest other point
--- a/test/trees/test_degenerate_extent.jl
+++ b/test/trees/test_degenerate_extent.jl
@@ -1,0 +1,117 @@
+using Test
+using ConservativeRegridding
+using ConservativeRegridding.Trees
+import GeometryOps as GO
+using GeometryOps.UnitSpherical: UnitSphericalPoint
+
+@testset "CellBasedGrid SphericalCap extent — degenerate cases" begin
+    manifold = GO.Spherical()
+
+    # Regression test: a (nlon+1, 2) grid spanning 180° of longitude across
+    # the equator produces corner points whose Cartesian sum is exactly
+    # (0, 0, 0). Before the fix, normalize((0,0,0)) → NaN poisoned the
+    # SphericalCap radius, silently breaking the dual DFS.
+    @testset "half-sphere equatorial band (symmetric cancellation)" begin
+        to_sphere = GO.UnitSphereFromGeographic()
+        nlon = 8  # 8 cells spanning 0°–180° (half the sphere)
+        dlon = 180.0 / nlon
+        pts = Matrix{UnitSphericalPoint{Float64}}(undef, nlon + 1, 2)
+        for k in 1:(nlon + 1)
+            lon = (k - 1) * dlon
+            pts[k, 1] = to_sphere((lon, -15.0))  # south face
+            pts[k, 2] = to_sphere((lon, +15.0))  # north face
+        end
+        grid = Trees.CellBasedGrid(manifold, pts)
+
+        # Full-range extent: all 8 cells
+        cap = Trees.cell_range_extent(grid, 1:nlon, 1:1)
+        @test isfinite(cap.radius)
+        @test cap.radius > 0
+
+        # The cap should cover roughly a hemisphere (180° span in lon)
+        @test cap.radius > π / 4   # at least 45°
+        @test cap.radius < π + 0.1 # at most slightly over π
+    end
+
+    @testset "full-sphere equatorial ring (360° longitude)" begin
+        to_sphere = GO.UnitSphereFromGeographic()
+        nlon = 16  # full ring
+        dlon = 360.0 / nlon
+        pts = Matrix{UnitSphericalPoint{Float64}}(undef, nlon + 1, 2)
+        for k in 1:(nlon + 1)
+            lon = (k - 1) * dlon
+            pts[k, 1] = to_sphere((lon, -15.0))
+            pts[k, 2] = to_sphere((lon, +15.0))
+        end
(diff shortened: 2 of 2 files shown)


src/trees/grids.jl
```diff
@@ -173,21 +173,58 @@ Since we know our grids are curvilinear, this is broadly okay.  It's not perfect
 =#
 
 using LinearAlgebra
+
+# Convert geographic (lon, lat) points to unit-sphere Cartesian, but pass
+# through points that are already UnitSphericalPoints (avoids a
+# double-conversion that produces NaN when CellBasedGrid stores
+# UnitSphericalPoint vertices directly).
+_to_unit_sphere(pt::GO.UnitSpherical.UnitSphericalPoint) = pt
+_to_unit_sphere(pt) = GO.UnitSphereFromGeographic()(pt)
```
asinghvi17: We have https://github.com/JuliaGeo/GeometryOps.jl/blob/c2a8837105f5bdd817c4e3ea5807259b7af7ba5f/src/utils/UnitSpherical/coordinate_transforms.jl#L57 in GeometryOps so I'm surprised this fixed anything new.  Are you sure something didn't become a regular 3-tuple or array somewhere?  Otherwise none of the existing regridding would work.

src/trees/grids.jl
```diff
@@ -173,21 +173,58 @@ Since we know our grids are curvilinear, this is broadly okay.  It's not perfect
...
+    function _safe_midpoint(a, b)
+        d = LinearAlgebra.dot(a, b)
+        if d > -0.985  # angle < ~170°, slerp is safe
+            return GO.UnitSpherical.slerp(a, b, 0.5)
+        else
+            mid = (a .+ b) ./ 2
+            n = LinearAlgebra.norm(mid)
+            return n < 1e-14 ? a : LinearAlgebra.normalize(mid)
+        end
+    end
```
asinghvi17: This doesn't need to be defined within the function, could it be pulled outside of it and given tests for the cases where numerical stability is effected (if you have e.g. a geojson of polygons, that could also be very useful)?

src/trees/grids.jl
```diff
@@ -173,21 +173,58 @@ Since we know our grids are curvilinear, this is broadly okay.  It's not perfect
...
-    p34 = GO.UnitSpherical.slerp(p3, p4, 0.5)
-    p41 = GO.UnitSpherical.slerp(p4, p1, 0.5)
+    raw_center = (p1 .+ p2 .+ p3 .+ p4) ./ 4
+    nrm = LinearAlgebra.norm(raw_center)
+    # When the four corners are (anti)symmetrically placed around the origin
+    # (e.g., a latitude band straddling the equator with exactly half the
+    # sphere in longitude), their sum cancels to near-zero. In that case,
+    # fall back to using ALL points (corners + borders) to find a non-
+    # degenerate centroid, or if that also cancels, use an arbitrary axis.
+    if nrm < 1e-12
```
asinghvi17, line 214: Is there a particular reason for 1e-12, or just that was what worked?  I guess the norm here is effectively an inverse measure of the angle between the original points.

It seems fine to me but mostly curious how you got to that specific number :)

asinghvi17 (commented):
Thanks for the PR!  A couple of comments / questions here:

asinghvi17:
To address the proposed other PRs:


1. **Missing `_intersects(Extent, SphericalCap)` method (in GeometryOps):** Generally the method to get a spherical extent is a bit different from that to get a regular extent, and a 2D Extent <-> SphericalCap comparison would be relatively expensive.  I would generally tend to prefer a 3D box in that case, but I don't think we have that yet - will probably put up a new package for flexible static R-trees soon.  

  That being said, I'm happy to have some method in GeometryOps to do that, as long as it's advised against - ideally the tree query should be relying on as many cached values as possible since it's very easy to bloat runtime there, hence also the CellBasedGrid storing values in UnitSphericalPoint.

2. **Built-in ESMF offline-weights export via NetCDF:** I'm not super clear why you'd want this (as opposed to just loading the weights from the other side into Julia), but this could certainly live in a NCDatasets.jl extension here.  

3. **Raw-corner cubed-sphere `treeify`:** A `treeify(Spherical(), corner_matrices::NTuple{6, Matrix{UnitSphericalPoint}})` method in the core package, or perhaps even making a separate submodule to hold useful tree structures, could definitely be good.

4. **JLD2 save/load example in the docs:** The `Regridder` struct is trivially serializable, but this isn't documented anywhere.  **Does it need to be?  I tend to avoid JLD2 because it'll break if you change type hierarchy or parameters.  Given the proposal (2) already I would prefer to go there instead**

5. **Benchmarks at realistic scale:** A table of (grid pair, nnz, build time, threads) in the docs would help users evaluating CR.jl for production grids (C180, C720, N320).  **Sure, would be great to have!  Does ESMF have a set of similar benchmarks by any chance?**

7. **`normalize=true` default + the TODO:** The default `normalize=true` divides by `max(intersections)`, making weight-level comparisons with xESMF impossible. The code has a `# TODO is this the best normalizer?` comment. At minimum the docstring should warn about this.  **Happy to have a warning about this but IIRC this is quite useful for the in-memory regridding work?  Perhaps @milankl should chime in here about what works best or if there is a better normalization approach**

8. **`areas()` ordering contract:** The implicit contract that `areas[k]` matches sparse-matrix index `k` is undocumented and fragile for composite trees. **Good point, should be documented for sure.  The actual contract is between `getcell(tree, i)` corresponding to `areas(tree)[i]`, which is also done in several other places, FWIW.**

9. **`RegularGrid{Spherical}` with lon-lat tuples:** It's unclear whether `GO.area(Spherical(), polygon_with_lonlat_tuples)` is correct, since the Oceananigans extension always converts to `UnitSphericalPoint` first.  **I think it should be but would have to check**

cfranken:
Thanks for the review @asinghvi17! Updated the PR based on your feedback:

1. **UnitSphereFromGeographic pass-through**: You're right, the pass-through at `coordinate_transforms.jl:57` already handles UnitSphericalPoints. I removed the redundant `_to_unit_sphere` helper entirely. The actual bugs were only the degenerate-center NaN and slerp instability.

2. **`_safe_slerp` extracted**: Pulled out of `circle_from_four_corners` into a module-level function with its own docstring.

3. **`1e-12` threshold**: It's ~`eps(Float64) * 1e4`. The corner vectors are O(1) (unit sphere), so cancellation from summing 4 of them produces residuals at the `eps` level. The 1e4 safety margin avoids false triggers from legitimate small-but-nonzero centroids (e.g., a very narrow lat band near the equator). Happy to adjust if you have a preferred value.

4. **Redundant `using LinearAlgebra`**: Removed.

5. **Regression test added**: `test/trees/test_degenerate_extent.jl` (27 tests) covering the half-sphere equatorial band (exact cancellation), full 360° ring, non-degenerate sanity, and an end-to-end regridder check confirming all 16 equatorial-ring cells find nonzero intersections.

Also noted in the commit message: the `RegularGrid{Spherical}` path with raw `(lon, lat)` tuple vertices is incompatible with `CellBasedGrid{Spherical}` with `UnitSphericalPoint` vertices — the clipping algorithm can't mix polygon vertex types. This is a separate issue (mentioned in the PR description as item 8) that we'd be happy to look into as a follow-up.

cfranken:
Thanks for the detailed response on the follow-up items!

**Re (2) ESMF export**: The main value is **cross-validation** — we diff CR.jl weights against xESMF at the individual `S[k]` level to verify polygon-intersection correctness (see the xESMF comparison test already in CR.jl). Having a native ESMF-format writer means you can do this without a Python round-trip. Secondary use: handing weights to ESMF-based Fortran tooling (GCHP, CESM) that consumes the standard `(S, row, col)` NetCDF format. PR #84 has the NCDatasets extension implementation.

**Re (4) JLD2 vs ESMF for persistence**: Fair point on JLD2 fragility across type hierarchy changes. Our use case is same-session caching (build weights once during preprocessing development, reload many times) where JLD2's speed matters and version stability doesn't. For long-term archival or cross-tool sharing, ESMF NetCDF is strictly better. We could document both with appropriate warnings — JLD2 for development cache, ESMF for archival/interop.

**Re (5) Benchmarks**: I don't know of a standard ESMF benchmark suite, but I can run CR.jl on a few representative grid pairs (LatLon 0.5° ↔ C48/C90, ReducedGaussian N80 ↔ C48, HEALPix nside=64 ↔ LatLon) and post a table. Would a PR to the docs be the right place?

**Re (7) normalize**: Makes sense to keep `normalize=true` as the default for in-memory work. A docstring note warning that `normalize=false` is needed for weight-level comparisons (xESMF parity, ESMF export) would be enough.

**Re (8) areas() ordering**: Good to know the contract is `getcell(tree, i) ↔ areas(tree)[i]` — a one-line docstring addition would save future users from having to discover this experimentally.

Happy to contribute PRs for any of these once #83 and #84 land. The code is already written and tested in our AtmosTransport.jl integration — it's mostly a matter of extracting and packaging it.

PR #84, opened 2026-04-12
Raw-corner cubed-sphere treeify + ESMF offline-weights export

cfranken:
## Summary

Two feature additions for users integrating CR.jl into climate model preprocessing pipelines. Developed while integrating CR.jl into [AtmosTransport.jl](https://github.com/RemoteSensingTools/AtmosTransport.git) for conservative regridding between LatLon, CubedSphere, and ReducedGaussian meshes.

**Note:** This PR includes the bug fix from #83 as a prerequisite (the raw-corner CS treeify uses `CellBasedGrid{Spherical}` with `UnitSphericalPoint` vertices, which triggers the `circle_from_four_corners` bugs fixed there). Once #83 is merged, this PR can be rebased to remove the duplicate commit.

## 1. Raw-corner cubed-sphere `treeify`

```julia
treeify(Spherical(), faces::NTuple{6, Matrix{UnitSphericalPoint}})
```

General-purpose cubed-sphere tree builder from 6 per-face `(Nc+1)×(Nc+1)` corner-point matrices — **no ClimaCore dependency required**.

Each face becomes a `CellBasedGrid` wrapped in `IndexOffsetQuadtreeCursor`, assembled into a `CubedSphereToplevelTree`. Global cell indexing is column-major per face, faces concatenated:
```
global_idx = (face-1) * Nc² + i + (j-1) * Nc
```

This produces the same tree structure as the ClimaCore extension, but accepts raw coordinate arrays from any source (GEOS-FP gnomonic projection, MPAS mesh files, user-defined grids, etc). The ClimaCore extension could be simplified to extract corners and delegate to this method.

**Use case:** GEOS-FP/FV3 cubed-sphere grids where you have the gnomonic corner coordinates but don't depend on ClimaCore.

## 2. ESMF offline-weights export

```julia
using NCDatasets
save_esmf_weights("weights.nc", regridder;
    src_grid_name="ERA5_0.25deg", dst_grid_name="C90")
```

Exports the regridder weights to the standard ESMF format consumed by **xESMF**, **ESMF_RegridWeightGen**, and **GCHP's tile-file loader**:

| Variable | Dim | Description |
|----------|-----|-------------|
| `S` | `(n_s,)` | Weight = `intersection_area / dst_cell_area` |
| `row` | `(n_s,)` | Destination cell index (1-based) |
| `col` | `(n_s,)` | Source cell index (1-based) |
| `frac_a` | `(n_a,)` | Source cell fraction covered |
| `frac_b` | `(n_b,)` | Destination cell fraction covered |
| `area_a` | `(n_a,)` | Source cell areas |
| `area_b` | `(n_b,)` | Destination cell areas |

Implemented as a **package extension** triggered by `using NCDatasets`. The function stub is always available for docstring visibility. Uses destarea normalization (the xESMF/ESMF default).

**Use case:** Cross-validate CR.jl weights against xESMF at weight level, or hand off weights to ESMF-based Fortran tooling.

## Relationship to #83

This PR builds on the bug fix in #83 (`circle_from_four_corners` UnitSphericalPoint handling). The raw-corner CS treeify and ReducedGaussian per-ring treeify both create `CellBasedGrid{Spherical}` trees with `UnitSphericalPoint` vertices, which would trigger the NaN extent bug without that fix.

## Testing

Tested via AtmosTransport.jl's `test_v2/regridding/runtests.jl` (564 tests, all passing), which exercises:
- LatLon ↔ CubedSphere conservation (constant, smooth fields, both directions)
- CubedSphere panel-center golden tests (Gnomonic + GEOS conventions)
- ESMF NetCDF round-trip (schema, frac_a/frac_b ≈ 1.0, index ranges)
- JLD2 serialization round-trip
- ReducedGaussianMesh ↔ LatLon/CubedSphere conservation

Suggesting also adding dedicated tests to CR.jl's test suite.

## Future PRs

As mentioned in #83, we're also happy to contribute:
- RingGrids extension implementation (per-ring `MultiTreeWrapper` for reduced Gaussian grids)
- Fix `RegularGrid{Spherical}` to use `UnitSphericalPoint` internally (vertex-type unification)
- Benchmarks at production scale


ext/ConservativeRegriddingNCDatasetsExt.jl
```diff
@@ -0,0 +1,128 @@
...
+Provides `save_esmf_weights(path, regridder; ...)` which writes the standard
+ESMF weight-file format (S, row, col, frac_a, frac_b, area_a, area_b).
+"""
+module ConservativeRegriddingNCDatasetsExt
+
+using ConservativeRegridding
+using ConservativeRegridding: Regridder
+using NCDatasets
+using SparseArrays: rowvals, nonzeros, nzrange
+using Dates: now
```
asinghvi17, line 14: Not sure we want to depend on Dates here, but if we do it'd have to be either a standard dependency, or an extra trigger for the extension.

PR #88, opened 2026-04-19
ESMF offline-weights export via NCDatasets extension

asinghvi17:
## Summary
Split from #84 — this is the ESMF export feature in isolation.

Adds `save_esmf_weights(path, regridder; kwargs...)` as a package extension triggered by `using NCDatasets`. Exports the regridder's sparse intersection matrix to the standard ESMF offline-weights NetCDF format (`S`, `row`, `col`, `frac_a`, `frac_b`, `area_a`, `area_b`) with destarea normalization.

This is the format consumed by **xESMF**, **ESMF_RegridWeightGen**, and **GCHP's tile-file loader**, enabling weight-level cross-validation between CR.jl and Python/Fortran tooling, or handoff to ESMF-based Fortran preprocessing pipelines.

The function stub is declared in the main module for docstring visibility; the implementation lives in `ext/ConservativeRegriddingNCDatasetsExt.jl` and is only loaded once `NCDatasets.jl` is in the environment.

Original author: @cfranken. Extracted from #84 to land atomically.

## Changes
- `Project.toml` — add `NCDatasets` as weak dep + extension registration
- `src/ConservativeRegridding.jl` — add `save_esmf_weights` function stub with docstring
- `ext/ConservativeRegriddingNCDatasetsExt.jl` (new, 128 lines) — extension implementing the ESMF writer

## Test plan
- [ ] Load with `using ConservativeRegridding, NCDatasets` and confirm the extension loads cleanly
- [ ] Call `save_esmf_weights` on a small regridder and validate NetCDF schema (variables, dims, fill values)
- [ ] Round-trip: read the NetCDF back and confirm `sum(S[row .== k])` ≈ `frac_b[k]` ≈ 1 for well-covered destination cells
- [ ] Cross-check a handful of weights against xESMF output on the same source/dest pair


ext/ConservativeRegriddingNCDatasetsExt.jl
```diff
@@ -0,0 +1,59 @@
...
+
+function ConservativeRegridding.save_esmf_weights(
+        path::AbstractString, r::Regridder;
+        src_grid_name::AbstractString = "source",
+        dst_grid_name::AbstractString = "destination",
+        src_shape::Union{Nothing, Tuple} = nothing,
+        dst_shape::Union{Nothing, Tuple} = nothing,
+        created_at::Union{Nothing, AbstractString} = nothing,
+    )
+    A         = r.intersections
```
simone-silvestri, line 25: If A is on a GPU we probably need to `Array` it? (also for src_areas and dst_areas?)

asinghvi17: Makes sense, implemented in https://github.com/JuliaGeo/ConservativeRegridding.jl/pull/88/commits/7af62377d146a519928345841eb9442c15729a4e

Issue #89, opened 2026-04-21
Implement treeify for SpeedyWeather grids


simone-silvestri:
In https://github.com/NumericalEarth/NumericalEarth.jl/pull/160 I am trying to switch from XESMF to ConservativeRegridding ad the regridding backend for NumericalEarth.jl.

I have noticed that we are missing a specialized `treeify` method for SpeedyWeather grids. 
I think it will not be too difficult because it should be very similar to what is already there for the `HealPix` grids.

I will open a PR soon with a proposal for building an efficient tree for SpeedyWeather native grids. I think we should target 3 main types:
- FullClenshawGrid (same as Oceananigans' LatitudeLongitudeGrid
- HealPixGrid
- OctaHEALPixGrid

Maybe @milankl do you have any suggestions for grids that need to be supported?

milankl:
The OctahedralGaussianGrid would be fantastic as it's our current default but the treeification is likely much harder as it's not a hierarchial grid as the others are ... Sorry I haven't found time for this despite talking with Anshul about it. The FullClenshawGrid is unravelled into a vector so you can just view it as a matrix and you're good to go. The HEALPix grids use ring order in SpeedyWeather and so I believe we'd need to let the treeify know about that ordering. We can transform to a nested order which means that always 4 consecutive points in the underlying vector belong together in the hierarchy, forming a cell one zoom level up. There's more on this here https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/887

asinghvi17:
We should be able to do the FullClenshawGrid pretty easily. 

For the implementation of HEALPIX and OctaHEALPIX, we would probably need a bit of clarification from SpeedyWeather around how to do the hierarchy and so on, but I think that should be doable, given that we already have Healpix.jl grids as a template.

Treeify needs to know about the ordering, but we don't necessarily need to switch the ordering of the data. We are already able to handle nested and ring order in Healpix.jl.  And we could perhaps just use the internal nqr (?) indexing that healpix already has, basically (face, i, j, level).

simone-silvestri:
Then a probably practical plan is to implement FullClenshawGrid first (which was the only supported by XESMF anyways) and move to HealPix and finally OctaHealPix

asinghvi17:
[...]

PR #90, opened 2026-04-23
Regridder implementation for SpeedyWeather's FullGrids

simone-silvestri:
This is a tentative PR to allow regridding to SpeedyWeather's `FullGrid`s, aka regular grids in LatitudeLongitude space (with the same number of longitude points per latitude ring).

I had a lot of assistance from Claude so I am not 100% I am right with the treeify, I will check a bit more thoroughly. 
However, tests pass (also the sweat test) provided we include SpeedyWeather's grid in the zoo of grids with the `is_dateline_straddling_source` problem.

Will post here some visuals of the regridding. 
Progress towards https://github.com/JuliaGeo/ConservativeRegridding.jl/issues/89


simone-silvestri:
BTW not sure if we want to keep the `SpeedyWeatherExt` or if we can just keep the `RingGridsExt` that fires with either `SpeedyWeather` and `RingGrids`

asinghvi17:
I think we could get rid of the SpeedyWeatherExt, since we only care about RingGrids types.   The treeify implementation looks good to me!

milankl:
Yes, given SpeedyWeather depends on RingGrids, only having `RingGridsExt` sounds good to me! Terrarium can also use RingGrids, so SpeedyWeather isn't needed!

PR #94, opened 2026-04-26
Add should_parallelize tree-level API for parallel task spawning

asinghvi17:
## Summary

Rewrite of #69 along the lines I think this should actually be designed.

The original PR fixed the planar-grid threading regression by replacing the internal `_area_criterion(extent::Extent)` `error()` with a permissive `true`. That makes the bug go away but bakes a single hard-coded multithreading policy into the package. This rewrite turns the criterion into a **public, tree-level API** instead, so each tree (built-in or user-defined) can supply its own granularity policy:

```julia
Trees.should_parallelize(tree, node, extent) -> Bool
```

Returning `true` means *"this node is the granularity at which we spawn a parallel task"* — the dual DFS stops descending single-threaded and runs the subtree as a single parallel unit.

## What's in the API

- `Trees.should_parallelize(tree, node, extent)` — extension point, exported from `Trees` and re-exported from `ConservativeRegridding`. Override per tree type to control multithreading granularity.
- `Trees.WithParallelizePolicy(tree, policy)` — instance-level wrapper for one-off tuning without subtyping. The callback signature is `policy(tree, node, extent) -> Bool`.

## Defaults shipped

- `extent::SphericalCap`: preserves the prior heuristic (cap < ¼ sphere). Works because spherical caps have a natural upper bound (the full unit sphere ≈ 4π sr) against which "small enough" has fixed meaning.
- `extent::Extents.Extent`: errors with a message pointing the tree author at the right method to define. **No planar default ships** — planar extents have no canonical scale, no upper or lower bound on magnitude, so no manifold-agnostic default can decide "small enough" without knowing the tree's own notion of size.

## Internals

- `MultithreadedDualDepthFirstSearch` now takes a pair of `(node, extent) -> Bool` closures (one per tree, pre-bound to its tree by `intersection_areas.jl`). Each node's extent is computed exactly once and threaded through the recursion — pulled out of the inner loop so child extents aren't recomputed per `(child1, child2)` pair either.
- `_area_criterion` removed entirely.

## Test plan

- [x] `test/regridding.jl`: planar threaded-regridding regression test now uses an in-test override on `TopDownQuadtreeCursor` to demonstrate and verify the dispatch hook
- [x] `test/regridding.jl`: new `WithParallelizePolicy wrapper` testset verifies the instance-level wrapper invokes its policy
- [x] Existing tests: `test/trees/grids.jl`, `test/trees/quadtree_cursors.jl`, `test/usecases/simple.jl` all pass
- [ ] Extension test suites (Oceananigans, ClimaCore, etc.) — spherical heuristic is unchanged so these should be unaffected, but worth verifying in CI

### DIFF
--- a/src/trees/interfaces.jl
+++ b/src/trees/interfaces.jl
@@ -15,7 +15,59 @@ thing that is returned implements the `SpatialTreeInterface` methods.
 import GeometryOpsCore as GOCore
 import GeometryOps as GO
 import GeometryOps: SpatialTreeInterface as STI
+import Extents
 import SortTileRecursiveTree # in order to implement the `getcell/ncell` interface
+
+"""
+    should_parallelize(tree, node, extent) -> Bool
+
+Decide whether to spawn a parallel task at `node` of `tree` during the
+multithreaded dual-tree traversal used to find candidate intersecting
+cell pairs.
+
+Returning `true` means *this node is the granularity at which a task is
+spawned*: the dual DFS stops descending recursively into children for
+the purpose of further parallelism and runs the subtree as a single
+parallel unit. Returning `false` means *keep descending* — the node is
+still too coarse and a task at this level would create unbalanced work.
+
+`extent` is the bounding region of `node` (e.g. an `Extents.Extent` for
+planar grids or a `SphericalCap` for spherical grids). It is passed
+explicitly so implementations and the dual DFS share one computation per
+node.
+
+# Defaults
+
+- `extent::SphericalCap`: spawns once the cap covers less than ¼ of the
+  unit sphere — a heuristic that avoids spawning a single task at the
+  root.
+- `extent::Extents.Extent`: errors. Planar trees must define a method
+  specific to their tree type, e.g.
+  `Trees.should_parallelize(::MyTree, node, extent::Extents.Extent) = ...`,
+  because there is no manifold-agnostic notion of "small enough" for
+  planar extents.
+
+# Customization
+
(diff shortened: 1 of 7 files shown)


asinghvi17:
cc @ctessum

simone-silvestri (approved):

src/trees/interfaces.jl
```diff
@@ -15,8 +15,70 @@ thing that is returned implements the `SpatialTreeInterface` methods.
...
+function should_parallelize end
+
+should_parallelize(tree, node, extent::GO.UnitSpherical.SphericalCap) =
+    (2π * (1 - cos(extent.radius))) < π
+
+should_parallelize(tree, node, extent::Extents.Extent) = error(
+    "`Trees.should_parallelize` has no method for planar `Extents.Extent` on tree of type `$(typeof(tree))`. " *
+    "Planar trees must define a tree-type-specific method, e.g. " *
+    "`ConservativeRegridding.Trees.should_parallelize(::$(typeof(tree)), node, extent::Extents.Extent) = ...` " *
+    "to control multithreading granularity."
```
asinghvi17:
```suggestion
    """
    `Trees.should_parallelize` has no method for planar `Extents.Extent` on tree of type `$(typeof(tree))`.
    Planar trees must define a tree-type-specific method, e.g. 
    `ConservativeRegridding.Trees.should_parallelize(::$(typeof(tree)), node, extent::Extents.Extent) = ...`
    to control multithreading granularity.
    """
```

Issue #100, opened 2026-05-06
Introduce an `AbstractMapping` property to the `Regridder` object


simone-silvestri:
Heads up: this design is already implemented in #99 on the `ss-js/update-spectral` branch, I'm opening this issue to discuss if this is the design we actually want. If "yes" I will open a separate PR that generalizes the `Regridder` object before we merge #99.

In the previously developed #82 we came up with three concrete regridder structs (`Regridder`, `SEtoFVRegridder`, `FVtoSERegridder`) which share most of their fields and differ only in which area data they carry and what `regrid!` does after `mul!`.  The three regridder structs share `weight_matrix`, `dst_temp`, and `src_temp`, and differ only in:

- whether `dst_areas` is present (FV destination: yes, SE destination: no),
- whether `src_areas` is present (FV to FV needs it for `transpose`, others don't),
- what `regrid!` does after `mul!` (divide by `dst_areas`, or nothing).

With this design, adding a new regridding direction (e.g. SE to SE) requires a new struct, a new `Base.size`, and copies of every `regrid!` dispatch variant, but most of that underlying code is the same.

In #99 we collapsed all these types into a single parameterized `Regridder{W, M, V}` whose mapping type `M <: AbstractMapping` encodes the regridding direction. The code follows this logic:

```julia
abstract type AbstractMapping end
abstract type AbstractNormalizingMapping <: AbstractMapping end

struct FVtoFV{V <: AbstractVector} <: AbstractNormalizingMapping
    dst_areas :: V
    src_areas :: V
end

struct SEtoFV{V <: AbstractVector} <: AbstractNormalizingMapping
    dst_areas :: V
end
struct FVtoSE <: AbstractMapping end   # M⁻¹ baked in; no post-mul! step

struct Regridder{W, M <: AbstractMapping, V} <: AbstractRegridder
    weight_matrix :: W
    mapping       :: M
    dst_temp      :: V
    src_temp      :: V
end

# Familiar names preserved as type aliases
const FVtoFVRegridder{W, V} = Regridder{W, <:FVtoFV, V}
const SEtoFVRegridder{W, V} = Regridder{W, <:SEtoFV, V}
const FVtoSERegridder{W, V} = Regridder{W, <:FVtoSE, V}
```

`regrid!` collapses to two methods covering all current and foreseeable mappings:

```julia
# Default: just matvec — catches FVtoSE and any future non-normalizing mapping.
[...]

PR #101, opened 2026-05-08
Bugfix regridding onto a `RightCenterFolded` tripolar grid

simone-silvestri:
Another fix for the horrible case of the `RightCenterFolded`. A `RightCenterFolded` grid has duplicate slots in its fold row: cell at position i and cell at position Nx+1-i are the same physical cell, so the same value is stored in both cells and kept in sync by the halo machinery.

When the regridder builds the intersection matrix, it picks _ONE_ of the two slots as "primary" (with the actual polygon) and the other as "partner"  with a degenerate zero-area ghost polygon. Basically in the matrix the primary slot column holds all the area and the partner (the repeated half row) has all zeros

This is fine for forward regridding out of the tripolar grid, but when regridding backwards (towards the `RightCenterFolded`), 
we divide by `dst_areas` so we get into a `0/0 = NaN` situation.

MWE:
```julia
dst = Field{Center, Center, Nothing}(TripolarGrid(size = (20, 20, 1)))
src = Field{Center, Center, Nothing}(LatitudeLongitudeGrid(size = (20, 20, 1), latitude=(-80, 90), longitude=(0, 360), z=(0, 1)))
set!(src, (x, y) -> x)
regridder = ConservativeRegridding.Regridder(dst.grid, src.grid)
ConservativeRegridding.regrid!(dst, regridder, src)
ConservativeRegridding.regrid!(vec(interior(dst)), regridder, vec(interior(src)))
heatmap(interior(dst, :, :, 1))
```
image

The fix in this PR consists in splitting each fold cell's intersection equally between the two repeated cells. Each matrix entry then has half the intersection in the matrix _AND_ half the cell area in dst_areas. Both cells end up with the same regridded value and conservation is preserved because primary + partner = full.


asinghvi17:
Yikes!  Do the values of each grid cell get added during computation?  I guess if this is OK with the downstream consumers of this grid I don't have a problem with it...but ouch!

simone-silvestri:
well in oceananigans when we `sum(field)` we actually always throw away the "repeated" half grid, but the areas in oceananigans remain correct, they are only divided by 2 in the regridder.
Do you think there might be problems downstream?

simone-silvestri:
The problem here is that `dst_areas` will not necessarily represent a `dst_area` anymore but a more general "normalization", but in the framework of #100 where we generalize the regridder finalization it might not matter too much

asinghvi17:
Do you think this should be part of the regridder finalization instead?  Then you can just copy the values over to the fake cells, if you want, and avoid introducing the restructuring into `treeify`.  I'll plot out that interface shortly

simone-silvestri:
but this is what this PR is doing right? We are building a regridder without touching treeify and then correcting the dst_areas and the intersection areas to halve the last row only after the regridder has been built.

asinghvi17:
Right, sorry I skimmed the pr and thought there was some code in treeify, but evidently not :D 

Yeah I think it could be good to just bite the bullet on that, I'll implement the API in a separate PR and then we can go from there.

That being said, do you want to halve the values and divide across the two rows, or just duplicate?  We could add a postprocessing step in the tests that zeros out the duplicate row 😅 or the API function responsible could do so

simone-silvestri:
That is also an option, maybe easier. To tailor the `finalize!` to substitute the NaNs in the last half row with valid values, as we do in the fill_halo_regions! of oceananigans. Thinking about it, it might even be a better approach because in this way we are sure that the final row actually matches.

ext/ConservativeRegriddingOceananigansExt.jl
```diff
@@ -174,6 +183,60 @@ function Trees.treeify(
...
+    _mirror_fold_partners!(dst_arraylike, field.grid)
+    copyto!(vec(interior(field)), dst_arraylike)
+    return field
+end
+
+# Partner slots in a folded grid have zero rows in the intersection matrix, so
+# after mul!/normalize they hold 0 (or 0/0 = NaN).  Copy each primary's value
+# into its partner so the field shows the same physical value on both copies.
+_mirror_fold_partners!(_, _) = nothing
+function _mirror_fold_partners!(dst_arraylike, grid::RightCenterFoldedGrid)
```
simone-silvestri: this function might not be GPU-proof. I ll add a kernel

Project.toml
```diff
@@ -35,7 +36,7 @@ ConservativeRegriddingClimaCoreExt = "ClimaCore"
 ConservativeRegriddingHealpixExt = "Healpix"
 ConservativeRegriddingInterfacesExt = "Interfaces"
 ConservativeRegriddingNCDatasetsExt = "NCDatasets"
-ConservativeRegriddingOceananigansExt = "Oceananigans"
+ConservativeRegriddingOceananigansExt = ["KernelAbstractions", "Oceananigans"]
```
asinghvi17, line 39: Does Oceananigans always bring in KernelAbstractions?

simone-silvestri: yeah it should

asinghvi17: sounds good then

PR #102, opened 2026-05-09
Split regrid api into five functions for easier management

asinghvi17:


src/regridder/regrid.jl
```diff
@@ -27,7 +27,32 @@ aᵈ = sum(A; dims=2)
...
+function perform_regridding!(dst_field, regridder, src_field; kwargs...)
+    error(
+        """
+        `perform_regridding!` must be implemented for destination type $(typeof(dst_field)),
+        regridder type $(typeof(regridder)), and source type $(typeof(src_field)).
+        """
+    )
+end
+
+function finalize_regridding!(dst_field, regridder, src_field; kwargs...)
```
simone-silvestri: Shall we divide by areas in this `finalize_regridding!`?

simone-silvestri (approved):

asinghvi17:
@simone-silvestri @juliasloan25 I think I was wrong, when I said on Thursday that it was sufficient for `initialize_regridding!` to take only the source field and regridder, and `finalize_regridding!` to take only the destination field and regridder.

The case that breaks this is when you have one vector be a non-`DenseArray` and then the other vector be a `DenseArray`.  Consider the following:
* `src::DenseVector, dst::NotDenseVector`: you should do `mul!(regridder.dst_temp, regridder.intersections, src)` and then load `regridder.dst_temp` into `dst`.
* `src::NotDenseVector, dst::DenseVector`: you should load `src` into `regridder.src_temp`, then call `mul!(dst, regridder, src_temp)`.  Then you are done.
* `src::DenseVector, dst::DenseVector`: there should be no loading of anything, just a straight `mul!`.

Enabling the fast path for `::DenseVector, ::DenseVector` then requires that you are aware of the natures of both `src` and `dst`.

And this is not even getting started on what happens when you have multidimensional arrays.

***

I think the solution to this is to have additional extractor functions that extract the vectors you want to do the regridding on.  Those can then be passed to regrid.  The new api surface is then:
```
extract_source_vector(src, regridder) -> something
extract_dest_vector(dst, regridder) -> something
```
where `something` is anything that works with `mul!`.  Then you can do:

```julia
function regrid!(dst, regridder, src)
    src_arraylike = extract_source_vector(src, regridder)
    dst_arraylike = extract_dest_vector(dst, regridder)
    initialize_regridding!(regridder, src, src_arraylike)
    perform_regridding!(dst_arraylike, regridder, src_arraylike)
    finalize_regridding!(dst, regridder, dst_arraylike)
end
```

This way, `perform_regridding!` can be assured what it's working on, and `finalize_regrid!` then has to know what the destination is.
In this case I'm not sure what the use of `initialize_regrid!` is.  But it's probably fine to keep it.  Perhaps the split is that `extract_source_vector` only gets you the pointer in some sense, and `initialize_regrid!` actually loads data into it - if necessary.

In this case the dispatch limitations are, then, the following (where a type annotation `::T` refers to something that you can dispatch on):

```julia
extract_source_arraylike(src::T, regridder)::AbstractArray
extract_dest_arraylike(dst::T, regridder)::AbstractArray
initialize_regridding!(regridder, src::T, src_arraylike) # - you should be assured what your src_arraylike is, but **must** always define `T`
finalize_regridding!(dst::T, regridder, dst_arraylike) # - you should be assured what your dst_arraylike is, but **must** always define `T`
perform_regridding!(dst_arraylike::T1, regridder, src_arraylike::T2) # - fully specified T1 and T2 only, no ambiguities allowed!
```

Does this make sense?  I'll push it shortly and I guess we will see.

src/regridder/regrid.jl
asinghvi17: Yeah, that's the place.  The new API helps with this quite a bit.

simone-silvestri:
I think this looks great.

simone-silvestri (approved):

juliasloan25 (approved):

PR #111, opened 2026-06-04
Benchmark suite: grid-scaling + ConservativeRegridding-vs-XESMF construction, with PR-vs-base CI

asinghvi17:
Adds regridder **construction-time** benchmarks and a label-gated CI job that graphs PR-vs-base construction time on the PR, modeled on Makie's benchmark workflow. Construction time is the focus (not regrid throughput).

## What's here (`bench/`)
- **Grid-scaling** (`scaling.jl`) — construction time vs grid **size** for the two main families, Oceananigans LatLon and Healpix; one curve per family. Conda-free.
- **CR vs XESMF** (`xesmf.jl`) — construction time vs XESMF/ESMF on Oceananigans grids (XESMF only supports Oceananigans). Times CR **threaded**, CR **serial**, and XESMF so the comparison isn't a misleading N-threads-vs-1 chart.
- **Plotting** (`plots.jl`) + local entry point (`run.jl`) → graphs in `bench/results/`.
- **Shared grid factories** (`grids.jl`): coarse→fine pairs (source at half the destination resolution), swept by size tier.

## PR-vs-base CI (`.github/workflows/benchmark.yml`)
Label-gated: add the **`run benchmarks`** label (or use *Run workflow*). The job:
1. `compare.jl` benchmarks the **PR** and the **base** ref — the driver always comes from the PR checkout, and only the `ConservativeRegridding` package is swapped (base via a detached git worktree; each ref runs in an isolated subprocess whose loaded package is provable via a `pathof` sentinel). Results are primitive-only `NamedTuple`s, so they cross the package-version boundary cleanly.
2. Runs CR-vs-XESMF on the PR ref (conda; isolated with `continue-on-error`).
3. Pushes the graphs to an orphan **`benchmark-assets`** branch and embeds them inline in a single **sticky PR comment**, using only the stock `GITHUB_TOKEN` — no bot PAT or gist. Full-res graphs also upload as a workflow artifact.

## Sample local results (4 Julia threads, cap 50k cells)

**Construction scaling** (lower is better) and **CR vs XESMF** — graphs will post to this PR once the `run benchmarks` label is added.

ConservativeRegridding vs XESMF (Oceananigans):

| dst cells | CR (threaded) | CR (serial) | XESMF | CR/XESMF |
|--:|--:|--:|--:|--:|
| 512 | 1.9 ms | 6.1 ms | 6.1 ms | 0.31× |
| 2048 | 8.5 ms | 27.2 ms | 17.3 ms | 0.49× |
| 8192 | 35.0 ms | 123.8 ms | 63.7 ms | 0.55× |
| 32768 | 160.6 ms | 535.2 ms | 260.2 ms | 0.62× |

(Threaded CR is ~1.6–3.2× faster than ESMF weight-gen; single-threaded, ESMF's Fortran path is faster than CR's serial assembly at scale.)

## Notes
- **Stacked on #110** (`as/parallelize-sparsematrix`) so this PR's diff is benchmark-only. Can be retargeted to `main`.
- Draft until the CI run is confirmed green on this PR.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,103 @@
+name: Benchmarks
+
+# Regridder-construction benchmarks, modeled on Makie's PR-vs-base benchmark CI: the benchmark
+# DRIVER always comes from the PR checkout, and only the ConservativeRegridding package is swapped
+# between the PR ref and the base ref (each run in an isolated subprocess). Results are rendered to
+# graphs that are pushed to an orphan `benchmark-assets` branch and embedded inline in a sticky PR
+# comment. Heavy/label-gated: it runs only when a PR carries the `run benchmarks` label (add it to
+# trigger a run) or on manual dispatch.
+
+on:
+  pull_request:
+    types: [labeled, synchronize, reopened, opened]
+  workflow_dispatch:
+    inputs:
+      base_ref:
+        description: "Base git ref to compare against"
+        required: false
+        default: "main"
+      maxcells:
+        description: "Cap on destination cells per tier (bounds runtime)"
+        required: false
+        default: "50000"
+
+concurrency:
+  group: benchmarks-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark:
+    name: Regridder construction benchmarks
+    if: >-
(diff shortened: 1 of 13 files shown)


asinghvi17:
The actual benchmarks here are crap, but we'll improve on those shortly.  This is mainly for the setup and understanding.

PR #112, opened 2026-06-04
Trig-free spherical broad-phase extents (~1.6× faster construction)

asinghvi17:
## What

Make the broad-phase bounding-cap computation for spherical grids **trig-free**. `_spherical_cap` (curvilinear grids: Oceananigans, RegularGrid) and the `ExplicitPolygonGrid{<:Spherical}` extent (Healpix/explicit polygons) previously built each cap with **4 `slerp`s + per-point `spherical_distance`** — all `atan`/`cos`/`sqrt`.

## Why

Profiling `Regridder` construction (see `bench/construction-performance.md` on the benchmark branch #111) showed **candidate search was ~55% of construction time**, and its #1 hotspot was `_spherical_cap`'s trig. Two exact identities remove almost all of it:

- The **great-circle edge midpoint** at `t=0.5` is exactly `normalize(a + b)` — no `slerp`. (Verified equal to ~15 digits.)
- The **cap radius** = max great-circle angle = `acos` of the **min cosine** → dot products + a single `acos`, instead of one `spherical_distance` per sample point.

A `_midcos` guard handles the one degenerate case: on a large index-rectangle an edge can span ~180°, making its corners antipodal so `normalize(a+b)` is `NaN`. There the two corners already force the cap wide, so the midpoint is simply dropped. (GeometryOps' `slerp` returned a finite arbitrary point here; the explicit guard is more principled.)

## Results

Oceananigans `Ny=128` (32,768 dst cells), local (proportions are platform-independent):

| | candidate search | total construction |
| --- | --- | --- |
| threaded (4 thr) | 74.2 → **18.5 ms** (4.0×) | 143.8 → **89.9 ms** (1.6×) |
| serial | 277.6 → **64.1 ms** (4.3×) | 500 → **279.6 ms** (1.8×) |

Candidate search drops from ~55% to ~21% of construction — assembly (clipping + spherical area) is now the bottleneck.

## Correctness

- **Candidate count (274,688) and nnz (48,640) are byte-identical** before/after — the broad phase is provably unchanged.
- The full test suite (**2636 tests**) passes unchanged, including the `cell_range_extent spherical: bounding cap` characterization test, dual-DFS tests, and regridding-conservation use cases.

Stacked on `as/parallelize-sparsematrix` (#110). Investigation write-up + the `bench/profile_cr.jl` profiler live on the benchmark branch (#111).

### DIFF
--- a/src/trees/grids.jl
+++ b/src/trees/grids.jl
@@ -141,7 +141,8 @@ function cell_range_extent(q::ExplicitPolygonGrid{<: GO.Spherical}, irange::Unit
     end
     isempty(all_points) && return GO.UnitSpherical.SphericalCap(GO.UnitSpherical.UnitSphericalPoint(0.0, 0.0, 1.0), 0.0)
     center = LinearAlgebra.normalize(sum(all_points) / length(all_points))
-    radius = maximum(p -> GO.spherical_distance(center, p), all_points)
+    # Max great-circle angle = acos(min cosine): dot products + one acos, no per-point trig.
+    radius = acos(clamp(minimum(p -> LinearAlgebra.dot(center, p), all_points), -1.0, 1.0))
     return GO.UnitSpherical.SphericalCap(center, radius * 1.0001)
 end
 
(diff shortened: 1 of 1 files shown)


simone-silvestri (approved):
nice!

PR #117, opened 2026-06-11
Use GeometryOps' SutherlandHodgmanCache in spherical intersection assembly

asinghvi17:
Wires the Sutherland-Hodgman allocation cache from [GeometryOps#as/cache-sutherlandhodgman](https://github.com/JuliaGeo/GeometryOps.jl/tree/as/cache-sutherlandhodgman) into the regridder-construction hot loop, to characterize its performance impact.

## What

- New interface seam **`task_local_operator(op)`** (`@public`), called once per chunk in the parallel COO assembly so operators holding mutable per-task state (allocation caches) can hand each task a private copy.
- `DefaultIntersectionOperator` gains a `cache` field; on `Spherical` the hook attaches a fresh `SutherlandHodgmanCache` per task. Caches are not task-safe, hence per-task copies.
- Dispatch on the cache value (`::Nothing` vs cache) keeps the operator loadable against GeometryOps versions without the cache — it degrades to the existing no-cache path.
- The per-task operator is resolved behind a **function barrier**: its type may not be inferrable (depends on the loaded GeometryOps), so assembly pays one dynamic dispatch per chunk into a type-stable kernel, never one per candidate pair. (The naive version cost +2 allocs and a dynamic dispatch per pair — measurable.)
- `bench/compare.jl` **temporarily** pins GeometryOps to the cache branch for *both* refs, so the CI comparison isolates ConservativeRegridding's cache wiring. Revert to registered GeometryOps before merge.

### DIFF
--- a/src/regridder/regridder.jl
+++ b/src/regridder/regridder.jl
@@ -79,10 +79,16 @@ Default intersection operator for the given manifold.
 
 Implemented for `Planar` and `Spherical` manifolds at the moment.
 Will dispatch to the appropriate intersection operator / algorithm based on the manifold.
+
+On `Spherical`, when the loaded GeometryOps provides `SutherlandHodgmanCache`,
+[`task_local_operator`](@ref) hands each assembly task a copy carrying a private
+clipping-buffer cache (caches must not be shared across tasks).
 """
-struct DefaultIntersectionOperator{M}
+struct DefaultIntersectionOperator{M, C}
     manifold::M
+    cache::C
 end
+DefaultIntersectionOperator(manifold) = DefaultIntersectionOperator(manifold, nothing)
 
 function (op::DefaultIntersectionOperator{<: GeometryOps.Planar})(p1, p2)
     intersection_polys = #=try; =#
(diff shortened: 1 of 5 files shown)


src/regridder/regridder.jl
```diff
@@ -94,14 +100,28 @@ function (op::DefaultIntersectionOperator{<: GeometryOps.Planar})(p1, p2)
...
 
+# Dispatch on the cache so the no-cache path never passes a `cache` keyword —
+# keeping the operator loadable against GeometryOps versions without one.
+_sutherland_hodgman_intersection(alg, p1, p2, ::Nothing) =
+    GeometryOps.intersection(alg, p1, p2; target = GeoInterface.PolygonTrait())
+_sutherland_hodgman_intersection(alg, p1, p2, cache) =
+    GeometryOps.intersection(alg, p1, p2; target = GeoInterface.PolygonTrait(), cache)
+
+function task_local_operator(op::DefaultIntersectionOperator{<: GeometryOps.Spherical})
+    isdefined(GeometryOps, :SutherlandHodgmanCache) || return op
```
asinghvi17, line 120:
```suggestion
```

asinghvi17:
Closing as this functionality has been merged to main

PR #129, opened 2026-08-08
Let curvilinear grids carry their manifold through Regridder

asinghvi17:
## The problem

A grid built on a non-default sphere couldn't be handed to `Regridder`:

```julia
R_authalic = 6371007.180918474        # WGS84 equal-area radius
grid = CellBasedGrid(Spherical(; radius = R_authalic), points)
Regridder(grid, src)
#=> MethodError: no method matching best_manifold(::CellBasedGrid{…})
```

`ExplicitPolygonGrid`, `CellBasedGrid` and `RegularGrid` all store the manifold they were
built on and define `GOCore.manifold`, but nothing defined `best_manifold` for them, and
there is no catch-all `best_manifold` method to fall back on.

Adding only `best_manifold` would have moved the failure one step later: there was no
`treeify` method for `AbstractCurvilinearGrid` either, and a grid matches none of the
generic `treeify(manifold, grid)` branches — `isspatialtree` is `false`, it isn't an
`AbstractMatrix`, and `Base.isiterable` is `false` — so it falls through and returns
`nothing`.

## The change

Three methods and a dependency.

**`best_manifold(::AbstractCurvilinearGrid)`** returns the grid's own declaration.  Unlike a
bare point matrix, a grid has nothing to guess.

**`treeify(manifold, ::AbstractCurvilinearGrid)`** wraps the grid in a
`TopDownQuadtreeCursor`.  A `manifold` argument that differs from the grid's own overrides
it — an explicit override for development, testing, and users who want one — rebuilding the
grid with `ConstructionBase.setproperties`, sharing its geometry array rather than copying.
The rebuild is what keeps the tree self-consistent: `cell_range_extent` dispatches on the
grid's manifold *type parameter*, so a grid still declaring the old manifold would compute
node extents on a surface nobody asked for.

**`ConstructionBase` is now a dependency**, which is what lets that be one line.  The grid
structs needed no code to become compatible — they have plain default constructors, so the
default `setproperties` already rebuilds them — so `src/trees/grids.jl` is untouched by this
PR, and grid types outside this package are coercible for free as long as they have a
`manifold` field (the HEALPix and OctaHEALPix face grids in `ext/` do).

## Breaking change: no implicit manifold promotion

`Regridder(dst, src)` used to promote a `Planar`/`Spherical` mismatch to `Spherical()`.  It
no longer promotes anything: any mismatch throws an `ArgumentError` naming which manifold
came from each side and pointing at `Regridder(manifold, dst, src)`.

Two reasons this costs little in practice:

- The comparison was `== GO.Spherical()`, which recognizes only the default radius and
  collapsed any other sphere onto it — silently rescaling areas by `(R₁/R₂)²`.  Oceananigans
  already reports `Spherical(; radius = grid.radius)`.
- The pairing that promotion produced doesn't clip anyway, at any radius.  Handing a
  `(lons, lats)` grid and a `UnitSphericalPoint` grid to `Regridder(Spherical(), dst, src)`
  today throws `DimensionMismatch: No precise constructor for UnitSphericalPoint{Float64}
  found. Length of input was 2.` — so promotion was turning a clear manifold error into an
  obscure one downstream.

Callers pairing two grids that genuinely share a surface are unaffected; every mixed-manifold
call site in the test suite either pins matching radii or already passes the manifold
explicitly.

## Tests

40 new assertions in `test/trees/grids.jl`:

- `best_manifold` on all three grid types, asserting the one-argument constructor defaults
  are **unchanged**.
- `ConstructionBase.setproperties` round-trips each grid type onto a new manifold with
  geometry identity (`===`) preserved.
- `treeify` pass-through when the manifold matches, and override when it doesn't, for each
  grid type.
- End-to-end `Regridder` on the authalic sphere:
  `sum(intersections) ≈ sum(src_areas) ≈ sum(dst_areas) ≈ 4πR_A²`.  The authalic radius is
  1.6 m off `Spherical()`'s mean radius, so a radius that silently reverted to the default
  would show up here.
- Both a two-sphere mismatch and a planar/spherical pair are rejected, and
  `Regridder(manifold, dst, src)` gets you past it.

Full suite green on CI (Julia 1 and 1.11): **10022 passed, 0 failed**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,29 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- `GeometryOpsCore.best_manifold(::AbstractCurvilinearGrid)` returns the manifold the grid
-  was constructed with, and `Trees.treeify(manifold, ::AbstractCurvilinearGrid)` wraps an
-  already-constructed grid in a `TopDownQuadtreeCursor`. Together these let
-  `Regridder(dst, src)` accept a grid object directly and compute on the radius that grid
-  declares — e.g. `Regridder(CellBasedGrid(Spherical(; radius = R_authalic), points), src)`
-  — rather than having to restate the manifold in the three-argument constructor.
-  Where the manifold passed to `treeify` differs from the grid's own, the argument wins
-  and the grid is rebuilt onto it, sharing its geometry array rather than copying; the
-  grid's own declaration is the default that one-argument `treeify(grid)` resolves
-  through `best_manifold`. Guarding against an unintended surface stays with `Regridder`,
-  which compares its destination's manifold against its source's. Grid types outside this
-  package need a `Trees._rebuild_manifold` method to be coercible.
+- `best_manifold(::AbstractCurvilinearGrid)` and `Trees.treeify(manifold, ::AbstractCurvilinearGrid)`, so a grid can be passed straight to `Regridder(dst, src)` and computed on the manifold it declares — e.g. `Regridder(CellBasedGrid(Spherical(; radius = R_authalic), points), src)`.  A `manifold` passed to `treeify` overrides the grid's own, rebuilding it with ConstructionBase.
+- The grid types are ConstructionBase-compatible, so `setproperties(grid, (; manifold))` rebuilds a grid onto another manifold, sharing its geometry rather than copying.
 
 ### Changed
-- `Planar`/`Spherical` promotion in `Regridder(dst, src)` promotes to the spherical side's
-  own manifold, radius included, instead of only recognizing the default `Spherical()` and
-  collapsing to it. A planar grid paired with a sphere of non-default radius is therefore
-  no longer rejected as a manifold mismatch (whether the promoted pair can then be clipped
-  is unchanged — mixing lon/lat tuples with unit-sphere points is a separate limitation).
-- The "Destination and source manifolds must be the same" error now names which manifold
-  came from which side and explains how to declare one, since the common cause is a
-  *guessed* manifold (a bare point matrix defaults to `Spherical()`, i.e. the WGS84 mean
-  radius) meeting a declared one such as the authalic sphere.
+- **Breaking**: `Regridder(dst, src)` no longer promotes a `Planar`/`Spherical` mismatch to `Spherical` — any mismatch now throws, and the manifold to compute on must be passed explicitly as `Regridder(manifold, dst, src)`.
 
 ## v0.2.7
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,33 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- `GeometryOpsCore.best_manifold(::AbstractCurvilinearGrid)` returns the manifold the grid
+  was constructed with, and `Trees.treeify(manifold, ::AbstractCurvilinearGrid)` wraps an
+  already-constructed grid in a `TopDownQuadtreeCursor`. Together these let
+  `Regridder(dst, src)` accept a grid object directly and compute on the radius that grid
+  declares — e.g. `Regridder(CellBasedGrid(Spherical(; radius = R_authalic), points), src)`
+  — rather than having to restate the manifold in the three-argument constructor.
+  Where the manifold passed to `treeify` differs from the grid's own, the argument wins
+  and the grid is rebuilt onto it, sharing its geometry array rather than copying; the
+  grid's own declaration is the default that one-argument `treeify(grid)` resolves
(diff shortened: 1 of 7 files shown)


src/regridder/regridder.jl
```diff
@@ -66,13 +66,34 @@ function Regridder(dst, src; kwargs...)
...
-            GO.Spherical()
-        elseif dst_manifold == GO.Spherical() && src_manifold == GO.Planar()
-            GO.Spherical()
+        # Implicitly promote to spherical.  `Planar` on one side is the fallback for
+        # coordinates that declare no surface, not a claim that the grid is flat, so the
+        # `Spherical` side wins - and it wins whole, radius included.
+        if dst_manifold isa GO.Planar && src_manifold isa GO.Spherical
+            src_manifold
+        elseif dst_manifold isa GO.Spherical && src_manifold isa GO.Planar
+            dst_manifold
```
asinghvi17: I don't like implicit promotion, lets not do this and instead generally force the manifold to be passed explicitly.

src/trees/grids.jl
```diff
@@ -39,6 +39,8 @@ struct ExplicitPolygonGrid{M <: GOCore.Manifold, PolyMatrixType <: AbstractMatri
     polygons::PolyMatrixType
 end
 GOCore.manifold(grid::ExplicitPolygonGrid) = grid.manifold
+_rebuild_manifold(grid::ExplicitPolygonGrid, manifold::GOCore.Manifold) =
+    ExplicitPolygonGrid(manifold, grid.polygons)
```
asinghvi17: We should make these structs ConstructionsBase.jl compatible so we can use its rebuild or remake or whateveer the function is.

src/trees/interfaces.jl
```diff
@@ -166,6 +166,67 @@ and then you may also want to specialize on `STI.node_extent(::QuadtreeCursor{<:
...
+### Manifolds and `treeify` for grids
+
+A curvilinear grid *carries* the manifold it was built on (`GOCore.manifold`, defined per
+grid type in `grids.jl`), so unlike a bare matrix of points there is nothing to guess -
+`best_manifold` is simply that declaration.  This is what lets a grid built on a
+non-default sphere, e.g. `CellBasedGrid(Spherical(; radius = R), points)`, carry `R`
+through `Regridder(dst, src)`, which resolves its compute manifold via `best_manifold`.
+
+(These live here, rather than up with the other `treeify` methods, because they dispatch
+on `AbstractCurvilinearGrid` - which only exists from this point in the file onwards.)
```
asinghvi17: Succinctly describe behaviour so not tell the story.

src/trees/interfaces.jl
```diff
@@ -166,6 +166,67 @@ and then you may also want to specialize on `STI.node_extent(::QuadtreeCursor{<:
...
+
+The grid's declaration is therefore a *default*: it is what the one-argument
+`treeify(grid)` resolves through [`GOCore.best_manifold`](@ref) when no manifold is named.
+Guarding against an unintended surface belongs where two independently-derived manifolds
+meet - [`Regridder`](@ref) compares its destination's against its source's and refuses a
+mismatch - not here, where the caller has named one outright.
+"""
+function treeify(manifold::GOCore.Manifold, grid::AbstractCurvilinearGrid)
+    return TopDownQuadtreeCursor(_coerce_manifold(grid, manifold))
+end
```
asinghvi17: I don't think this function should exist.

src/trees/interfaces.jl
```diff
@@ -166,6 +166,67 @@ and then you may also want to specialize on `STI.node_extent(::QuadtreeCursor{<:
...
+`grid` rebuilt on `manifold`, sharing its geometry rather than copying it; `grid` itself
+when the manifold already matches, so the common path allocates nothing and preserves
+object identity.
+
+Each `AbstractCurvilinearGrid` needs a method: the geometry fields differ per type and
+there is no generic way to swap one field of an arbitrary struct without reaching for
+another dependency.
+"""
+_coerce_manifold(grid::AbstractCurvilinearGrid, manifold::GOCore.Manifold) =
+    GOCore.manifold(grid) === manifold ? grid : _rebuild_manifold(grid, manifold)
```
asinghvi17: this stuff will be replaced by the constructionsbase.jl things anyway

CHANGELOG.md
```diff
@@ -5,6 +5,33 @@ All notable changes to this project will be documented in this file.
...
+### Changed
+- `Planar`/`Spherical` promotion in `Regridder(dst, src)` promotes to the spherical side's
+  own manifold, radius included, instead of only recognizing the default `Spherical()` and
+  collapsing to it. A planar grid paired with a sphere of non-default radius is therefore
+  no longer rejected as a manifold mismatch (whether the promoted pair can then be clipped
+  is unchanged — mixing lon/lat tuples with unit-sphere points is a separate limitation).
+- The "Destination and source manifolds must be the same" error now names which manifold
+  came from which side and explains how to declare one, since the common cause is a
+  *guessed* manifold (a bare point matrix defaults to `Spherical()`, i.e. the WGS84 mean
+  radius) meeting a declared one such as the authalic sphere.
```
asinghvi17: decrease word vomit

PR #131, opened 2026-08-14
Use GeometryOps' SutherlandHodgmanCache in spherical intersection assembly

asinghvi17:
Supersedes #117 by porting the spherical intersection cache work onto the current `main` branch.

## What changed

- add a project-level `[sources]` entry that resolves GeometryOps from its `main` branch
- add the `task_local_operator` assembly hook and create a private `SutherlandHodgmanCache` for each chunk/task
- use that cache for spherical Sutherland–Hodgman intersections
- remove the optional `isdefined(GeometryOps, :SutherlandHodgmanCache)` fallback requested in the review comment, since the source override now guarantees the API
- add tests covering private cache creation and the unchanged planar fallback

## Impact

Spherical intersection assembly can reuse clipping buffers without sharing mutable cache state between tasks. Other intersection operators retain the existing behavior through the identity default for `task_local_operator`.

## Validation

- `git diff --check`
- clean-environment resolution confirmed GeometryOps v0.1.43 from `JuliaGeo/GeometryOps.jl#main`
- `test/regridding.jl` (64 tests passed)


src/regridder/intersection_areas.jl
```diff
@@ -70,6 +70,16 @@ By default, this is a no-op and returns the candidate pairs as is.
 """
 work_items(op, candidate_pairs) = candidate_pairs
 
+"""
+    task_local_operator(op) -> op′
+
+Return the operator instance one assembly task should use. Called once per chunk
+inside the parallel COO assembly (never in the per-item hot loop), so operators
+holding mutable per-task state — e.g. allocation caches — can hand each task a
+private copy. Defaults to returning `op` unchanged.
```
asinghvi17:
```suggestion
Return the operator instance which the current assembly task should use. 

This is mainly to enable intersection operators to materialize task-local caches.

Defaults to returning `op` unchanged.  For any threadsafe operator, this is a no-op.
```

Project.toml
```diff
@@ -23,6 +23,9 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StableTasks = "91464d47-22a1-43fe-8b7f-2d57ee82463f"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
+[sources]
+GeometryOps = {url = "https://github.com/JuliaGeo/GeometryOps.jl", rev = "main"}
+
```
asinghvi17:
```suggestion
```

PR #132, opened 2026-08-16
Fix threaded intersection query crashing on empty result sets

asinghvi17:
## What crashes

`intersection_areas(manifold, True(), dst_tree, src_tree)` throws

```
ArgumentError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer
```

whenever the two trees share **no** intersecting cell pair. The sequential path (`False()`) handles exactly the same input correctly, returning an all-zero `SparseMatrixCSC` of the right size and eltype.

https://claude.ai/code/session_01353CAguZrrCdbLC6iBT7Vq

### DIFF
--- a/src/regridder/intersection_areas.jl
+++ b/src/regridder/intersection_areas.jl
@@ -195,9 +195,11 @@ function assemble_sparse_matrix_coo(style::S, op::O, items::I, src_tree::T1, dst
     # Fetch the results of `result_tasks`
     all_results = map(fetch, result_tasks)
     # Concatenate the per-chunk COO vectors into single vectors, in partition order.
-    rows = reduce(vcat, getindex.(all_results, 1))
-    cols = reduce(vcat, getindex.(all_results, 2))
-    vals = reduce(vcat, getindex.(all_results, 3))
+    # `init` is load-bearing: there are no partitions when `items` is empty, which is
+    # what a candidate search that found no intersecting pair returns.
+    rows = reduce(vcat, getindex.(all_results, 1); init = Int[])
+    cols = reduce(vcat, getindex.(all_results, 2); init = Int[])
+    vals = reduce(vcat, getindex.(all_results, 3); init = ValType[])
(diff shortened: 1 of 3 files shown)


src/regridder/intersection_areas.jl
```diff
@@ -195,9 +195,11 @@ function assemble_sparse_matrix_coo(style::S, op::O, items::I, src_tree::T1, dst
     # Fetch the results of `result_tasks`
     all_results = map(fetch, result_tasks)
     # Concatenate the per-chunk COO vectors into single vectors, in partition order.
-    rows = reduce(vcat, getindex.(all_results, 1))
-    cols = reduce(vcat, getindex.(all_results, 2))
-    vals = reduce(vcat, getindex.(all_results, 3))
+    # `init` is load-bearing: there are no partitions when `items` is empty, which is
+    # what a candidate search that found no intersecting pair returns.
```
asinghvi17:
```suggestion
    # Because we insist on integer indices anyway, we can set init to an empty vector.
    # If the threaded search finds no candidate pairs, this prevents the dreaded `Reducing over an empty collection`
```

src/utils/MultithreadedDualDepthFirstSearch.jl
```diff
@@ -61,7 +61,9 @@ function multithreaded_dual_query(
         _inner_dfs_f, predicate, parallelize1, parallelize2, tasks,
         node1, ext1, node2, ext2,
     )
-    return reduce(vcat, map(fetch, tasks))
+    # `init` is load-bearing: `tasks` is empty whenever no child pair of the two
+    # roots satisfies the predicate, i.e. the trees do not intersect at all.
```
asinghvi17:
```suggestion
    # Because we insist on integer indices anyway, we can set init to an empty vector.
    # If the threaded search finds no candidate pairs, this prevents the dreaded `Reducing over an empty collection`
```

PR #136, opened 2026-08-19
Update area tests to include Planar parameter

asinghvi17:


asinghvi17:
Urgh, looks like I need a new release of GeometryOpsCore to establish Base.broadcastable for ::Manifold...

PR #137, opened 2026-08-19
Replace the threaded dual-tree spawn policy with a budget-frontier traversal

asinghvi17:
## Mechanism

The multithreaded dual depth-first search no longer spawns a task whenever both nodes' `should_parallelize` predicates fire. It now runs a **budget-frontier traversal**:

1. Descend both trees serially, keeping a max-priority frontier of node *pairs* keyed by estimated work (spherical-cap overlap area × cell density; a generic fallback covers other extent types). The heaviest pair is split repeatedly — one-sided splits when one side is a leaf; a pair is terminal only when both sides are leaves — until the frontier holds at least `chunks_per_thread × nthreads()` pairs *and* no pair claims more than an even `1/nchunks` share of the estimated work (hard stop 512 pairs past the budget, so a coarse estimate cannot run away in spawns). `chunks_per_thread` is a keyword on `multithreaded_dual_query`, default 8 (the measured winner).
2. Spawn one task per frontier pair, each running the serial `STI.dual_depth_first_search` on its subtree pair.
3. Concatenate the per-task results in DFS pre-order, reproducing the serial emission order bit-for-bit.

The recursive policy's failure was structural, not a tuning problem: its task count is fixed by tree shape (18,944–59,888 tasks on the benchmark workloads, up to 49 % of them empty, with single tasks holding up to 21 % of all pairs), and no `should_parallelize` spelling fixed all six workloads at once — the two hand-tuned variants measured each fixed some and regressed others. The frontier sizes tasks by measured weight, so the worst task share drops to 0.1–0.8 %.

## Measured verdict

Full campaign: 5 arms × 6 dual-tree workloads × t1/t8/t64 on a 64-core machine (min over 2 rounds × 3 reps), plus an end-to-end regrid. The frontier at `chunks_per_thread = 8` ("S2") vs the shipping policy ("A"):

**Traversal wall time, t64 (seconds; worst cases for the shipping policy):**

| workload | shipping | frontier | speedup |
|:--|--:|--:|--:|
| clima cubed-sphere → ocean lat-lon | 0.845 | 0.083 | 10.1× |
| ocean lat-lon → HEALPix nested | 1.178 | 0.186 | 6.3× |
| SpeedyWeather Clenshaw → Gaussian | 0.665 | 0.078 | 8.6× |
| ocean lat-lon → tripolar | 0.772 | 0.167 | 4.6× |

The shipping policy loses monotonicity at t64 on four of six workloads (t64 slower than its own t8, burning 12–28 CPU-s where the frontier burns 1.1–5.3). The frontier is fastest or tied-fastest on **all six workloads at t64**, within noise at t1 everywhere (all arms identical to <1 % serial), and within ~10 % at t8. Frontier planning cost: 4–22 ms.

**End-to-end regrid (GLO-30 DEM → DGGS via the lazy conservative path):**

| size | threads | shipping | frontier |
|--:|--:|--:|--:|
| 512² | 8 | 19.0 s | 17.3 s |
| 512² | 64 | 18.4 s | 16.3 s |
| 1024² | 8 | 30.2 s | 27.5 s |
| 1024² | 64 | 25.8 s | 22.2 s |

**Correctness:** the explicit gate (serial re-traversal, order and set compared element-wise) ran for all arms × 6 workloads × t1/t8/t64 — 55/55 identical, zero violations. End-to-end weight matrices agree (nnz, sum, sighash) across all arms, and all 30 end-to-end regrid runs return identical digests. The frontier reproduces the serial DFS emission order exactly, so downstream assembly is order-stable by construction. This PR adds a testset pinning that contract: `multithreaded_dual_query == dual_depth_first_search` (order and content) on a spherical and a planar grid pair, at `chunks_per_thread ∈ (1, 8, 64)` and through the back-compat signature, plus threaded `intersection_areas` `==` unthreaded.

The full test suite passes at 1 and 8 threads (10049/10049). At 64 threads the only failure is the pre-existing `test/extensions/climacore.jl:53` `isapprox(0.0, -0.75)`, which fails byte-identically on current `main` (it fires only at high thread counts).

## The even-share stop rule (second commit)

Gate-testing this PR on a plain 720×360 → 512×256 lat-lon pair found a case the campaign's six workloads missed: the bounding cap of a *quarter* of a global lat-lon grid is a whole hemisphere (or, southern quarters, the entire sphere), so eight of the sixteen root child pairs carry bit-identical weights while four of them hold ~333k candidate pairs each and four hold ~1000. A count-only stop then leaves single tasks holding 25 % of the traversal (t8: 1.52 s vs 0.86 s for the old policy). Two fixes, measured back to green:

- splitting continues past `nchunks` while the heaviest pair still exceeds its even share of the frontier's estimated work (bounded overshoot as above) — tied pairs all get split, and one level down the child extents separate real work from junk;
- caps with radius ≥ π — including the all-NaN cap GeometryOps returns for a point set with no usable centre, which any grid spanning both poles has at its root — are treated as the whole sphere (a NaN weight would sink in the max-heap and never be split).

After the fix: max task share 24.9 % → 5.3 %, t8 0.864 s (main) vs 0.871 s (branch) on that workload, and equal-to-better on the campaign-style workloads at t8 and t64 (e.g. Clenshaw→Gaussian t8 0.212→0.140 s). The pair vector stays byte-identical throughout — weights only order the split queue.

## Deprecations and removals

- `should_parallelize` and `WithParallelizePolicy` are **no longer consulted** — task granularity no longer depends on a per-node-type predicate. Both stay defined and exported so downstream method definitions and wrapped trees keep loading and working; they are slated for removal in a future breaking release.
- The default `should_parallelize` methods are removed (the `::Any` cap-area fallback, the planar `Extents.Extent` error method, and the quadtree-cursor defaults). A side effect: planar trees now thread out of the box instead of requiring the tree author to supply a policy (#66).
- The recursive `multithreaded_dual_depth_first_search` walker is removed. `multithreaded_dual_query` keeps a back-compat method that accepts (and ignores) the old parallelize closures.

## Extension point

`Trees.split_weight(node) -> Real`: an O(1) estimate of the work under a node, used only to order the split queue. The generic default answers via `ncells`, so most tree types need no method; extensions can add a one-liner (the HEALPix and RingGrids extensions do). A wrong answer can only cost load balance, never correctness.

https://claude.ai/code/session_01NYsLTQDR6sHuQGeHakPBT7


examples/multithreaded_dfs.jl
```diff
@@ -1,3 +1,10 @@
+# Historical prototype of the threaded dual-tree search: it spawns a task as soon as
+# both nodes' extents pass an area criterion.  ConservativeRegridding no longer works
+# this way - `multithreaded_dual_query` now splits the root node pair into a budget
+# frontier of `nthreads * chunks_per_thread` pairs (heaviest pair first, weighed by
+# `Trees.split_weight`) and spawns one task each, which also makes its output order
+# identical to the serial search.  Kept here as a benchmark reference point.
```
asinghvi17:
```suggestion
# Historical prototype of the threaded dual-tree search: it spawns a task as soon as
# both nodes' extents pass an area criterion.  ConservativeRegridding no longer works
# this way.  
# Its `multithreaded_dual_query` now splits the root node pair into a budget
# frontier of `nthreads * chunks_per_thread` pairs and spawns one task each.
```

src/trees/interfaces.jl
```diff
@@ -22,74 +22,44 @@ import SortTileRecursiveTree # in order to implement the `getcell/ncell` interfa
...
-
-2. **Instance-level wrapper** — wrap the root tree at construction time:
-
-   ```julia
-   tree_with_policy = WithParallelizePolicy(my_tree, (tree, node, extent) -> ...)
-   ```
-
-   The wrapper plumbs the policy through the dual DFS via a local closure
-   in `intersection_areas.jl`, so it doesn't participate in dispatch.
+!!! warning "Deprecated"
```
asinghvi17, line 25: Let's slap a depwarn on this

src/trees/interfaces.jl
```diff
@@ -22,74 +22,44 @@ import SortTileRecursiveTree # in order to implement the `getcell/ncell` interfa
...
+of grid cells beneath it.
+
+Used only to order the split heap that the multithreaded dual-tree traversal
+builds its task frontier from: a wrong answer costs load balance, never
+correctness.
+
+The generic method answers from [`Trees.ncells`](@ref) when the node type
+defines it, and returns `1` otherwise, so most tree types need no method at
+all. Define one for a node type whose `ncells` is missing or expensive, or
+whose cells are unevenly distributed across its subtrees.
```
asinghvi17:
```suggestion
A fast estimate of how many leaves live under `node`.

This is used to build a multithreading plan in `multithreaded_dual_dfs`.
A wrong answer here will impact runtime, but not correctness.

By default, this has a fallback that tries to check `Trees.ncells(node)`,
and returns `1` if that is not applicable for the node.
```

src/trees/interfaces.jl
```diff
@@ -22,74 +22,44 @@ import SortTileRecursiveTree # in order to implement the `getcell/ncell` interfa
...
+"""
+function split_weight(node)
```
asinghvi17:
```suggestion
function split_weight(node::N) where N
```

a little type specialization never hurt

src/trees/wrappers.jl
```diff
@@ -67,31 +68,26 @@ STI.node_extent(w::KnownFullSphereExtentWrapper) = GO.UnitSpherical.SphericalCap
 #=
 ## WithParallelizePolicy
 
-Wraps any spatial tree so its `should_parallelize` is supplied by a
-user-provided callable — the instance-level counterpart to defining a
-node-type-specific `should_parallelize` method.
+Deprecated: used to supply a tree's `should_parallelize` from a user callable.
```
asinghvi17, line 71: slap a depwarn here too

src/utils/MultithreadedDualDepthFirstSearch.jl
```diff
@@ -1,44 +1,195 @@
...
+surviving as a single task. Phase 2 spawns the tasks and concatenates their
+results in DFS pre-order, which makes the output bit-identical - and
+order-identical - to a full serial dual DFS.
+=#
+
+# A cap that covers the whole sphere, including the all-NaN cap GeometryOps
+# returns for a point set with no usable centre - a grid spanning both poles
+# has one. It localises nothing, and NaN must not reach the weights: a NaN
+# never compares greater, so a NaN-weighted pair sinks to the bottom of the
+# max-heap and is never split.
```
asinghvi17: reduce these comments please no one will ever read this much text

src/utils/MultithreadedDualDepthFirstSearch.jl
```diff
@@ -1,44 +1,195 @@
...
+subtree is handled by descending the large side only. Children failing
+`predicate` are dropped exactly as the serial dual DFS would prune them, so
+running the serial search on every returned pair visits each candidate pair
+exactly once.
+"""
+function frontier(predicate::P, root1, root2; nchunks::Int) where {P}
+    e1 = STI.node_extent(root1)
+    e2 = STI.node_extent(root2)
+    ws = Float64[]; xs = Any[]          # splittable pairs, as a max-heap
+    done = Any[]                        # leaf/leaf pairs, cannot split further
```
asinghvi17, line 118: We will eventually have to figure out some way to make this trim compatible.  But that day is not today.

src/utils/MultithreadedDualDepthFirstSearch.jl
```diff
@@ -49,22 +200,47 @@ function _inner_dfs_f(predicate::P, node1::N1, node2::N2) where {P, N1, N2}
...
+    predicate::P, node1::N1, node2::N2;
+    chunks_per_thread::Int = 8,
+) where {P, N1, N2}
+    nchunks = max(1, Threads.nthreads() * chunks_per_thread)
+    pairs = frontier(predicate, node1, node2; nchunks)
+    tasks = Any[]
+    sizehint!(tasks, length(pairs))
+    for p in pairs
+        push!(tasks, _spawn_pair(_inner_dfs_f, predicate, p[1], p[3]))
+    end
```
asinghvi17: Can't you just make this map(pairs) do pair and then spawn tasks in the map?  Seems a lot simpler.  And may get you some advantages from StableTasks.

CHANGELOG.md
```diff
@@ -8,14 +8,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- The multithreaded candidate search now plans its tasks with a *budget frontier*: a short serial pass splits the root node pair into at least `Threads.nthreads() * chunks_per_thread` independent node pairs, always splitting the pair with the largest estimated work and carrying on while any pair still looks larger than its even share of the total, and spawns one task each.  Results are concatenated in depth-first order, so the threaded search returns exactly the pairs the serial one does, in exactly the same order.
+- `multithreaded_dual_query(predicate, node1, node2; chunks_per_thread = 8)` exposes the task budget.  A higher budget tolerates more skew between node pairs at the cost of more spawns; it cannot change the result.
+- `Trees.split_weight(node) -> Int`, an O(1) estimate of the work under a node, used only to order the frontier's split queue - a wrong answer costs load balance, never correctness.  It answers from `Trees.ncells` by default, so most tree types need no method; the HEALPix and OctaHEALPix root nodes define their own.
```
asinghvi17: Please edit these to be more succinct as changelog entries.

CLAUDE.md
```diff
@@ -104,7 +104,7 @@ The manifold affects extent computation, intersection algorithms, and area calcu
 
 ### Multithreading
 
-`multithreaded_dual_query` (`src/utils/MultithreadedDualDepthFirstSearch.jl`): Parallel dual-tree traversal. Spawns tasks when both nodes are leaves or when nodes satisfy an area criterion (avoids spawning excessive tasks for small regions). Intersection area computation is separately parallelized via ChunkSplitters partitioning.
+`multithreaded_dual_query` (`src/utils/MultithreadedDualDepthFirstSearch.jl`): Parallel dual-tree traversal. A serial pass splits the root node pair into a frontier of `nthreads * chunks_per_thread` node pairs, always splitting the pair with the largest estimated work (`Trees.split_weight`), then spawns one task per frontier pair and concatenates the results in DFS order, so the output matches the serial traversal exactly. Intersection area computation is separately parallelized via ChunkSplitters partitioning.
```
asinghvi17: simplify this

examples/multithreaded_dfs.jl
asinghvi17: Applied verbatim.

src/trees/interfaces.jl
asinghvi17: Done - added a `Base.depwarn` fallback method next to `function should_parallelize end`:

```julia
function should_parallelize(node, extent)
    Base.depwarn("`Trees.should_parallelize` is no longer consulted by the multithreaded traversal - define `Trees.split_weight` instead.", :should_parallelize)
    return false
end
```

It's a fallback, so a tree author's own method still wins and stays silent; only calls that would previously have hit a default now warn. Verified firing under `--depwarn=yes`. Also trimmed the docstring while I was in there.

src/trees/interfaces.jl
asinghvi17: Applied, with one word changed: the function is `multithreaded_dual_query`, not `multithreaded_dual_dfs`, so I used the real name to keep the `@ref`-able text honest.

src/trees/interfaces.jl
asinghvi17: Done.

src/trees/wrappers.jl
asinghvi17: Done - the depwarn goes in an inner constructor so it fires on construction rather than on use (the policy is never used any more, so construction is the only hook left):

```julia
function WithParallelizePolicy(tree::T, policy::F) where {T, F}
    Base.depwarn("`WithParallelizePolicy` is deprecated - its policy is ignored, so wrapping is a no-op.  Define `Trees.split_weight` to tune task balance.", :WithParallelizePolicy)
    return new{T, F}(tree, policy)
end
```

One knock-on: defining an inner constructor drops the auto-generated `WithParallelizePolicy{T,F}(tree, policy)` form with explicit type parameters. Nothing in the repo used it, and `ConstructionBase.setproperties` goes through the un-parameterised call, so the wrapper still rebuilds fine.

src/utils/MultithreadedDualDepthFirstSearch.jl
asinghvi17: Cut. The module header is down from 15 lines to 4, and I gave the rest of the file the same treatment - the `_is_global` note, `pair_weight`, `MAX_EXTRA_PAIRS`, the `frontier` docstring, and the `multithreaded_dual_query` docstring. About 50 lines of prose gone; the file is 92 lines lighter overall.

src/utils/MultithreadedDualDepthFirstSearch.jl
asinghvi17: Agreed, not today. I left a one-line marker on the declaration so the reason is findable:

```julia
# `Any` because the pairs are heterogeneously typed - not trim-compatible yet.
```

Worth noting the cost is compile-time only: these vectors live entirely in the serial planning phase, which measured 4-22 ms across the sweep, so concretising them would buy nothing at runtime. It's purely a `--trim` blocker.

src/utils/MultithreadedDualDepthFirstSearch.jl
asinghvi17: Done, and you were right about StableTasks:

```julia
tasks = map(pairs) do pair
    _spawn_pair(_inner_dfs_f, predicate, pair[1], pair[3])
end
```

`_spawn_pair` is a function barrier that returns the same `StableTask{Vector{Tuple{Int,Int}}}` for every pair, so `map` infers it and `tasks` comes out as `Vector{StableTasks.StableTask{Vector{Tuple{Int, Int}}}}` instead of the `Vector{Any}` the `push!` loop built. The `fetch` loop is concrete now. Verified on a 49-pair frontier.

CHANGELOG.md
asinghvi17: Tightened these three, and trimmed the matching Changed/Removed/Fixed entries in the same PR so they read consistently.

CLAUDE.md
asinghvi17: Simplified.

asinghvi17:
**Follow-up: the frontier's work estimator was flat on raster → DGGS builds; fixed by conserving the estimate across splits.**

**Defect.** On an eager `RasterGrid(512²)` → IGEO7 level-13 plan, every frontier pair's estimated weight was nearly identical (CV 0.02) while actual pairs-per-task spanned five orders of magnitude; one task held 50 % of all 1.89 M candidate pairs and the t8 plan ran at utilization 2.8 instead of ~7. Root cause: the raster tree answers `Trees.ncells` — and therefore the `split_weight` fallback — with its **whole space** at every node, so its density estimate in `pair_weight` scales as `1/cap_area(node)`, and multiplying by the overlap area cancels exactly: any raster window facing a containing subtree weighs ≈ `N_total`. Splits never reduced the heaviest weight, each split *inflated* the estimated total, and the share test declared balance spuriously.

**Fix** (`frontier`): a split may only redistribute its parent's weight — surviving children whose raw weights sum past the parent's are scaled back to it, never up. Honest, subadditive estimates pass through unchanged (sibling ratios preserved, so the d4ef556 tied-pair descent is intact); a saturating estimate now degrades to an even breadth-first split instead of stalling. A regression test drives the frontier through a wrapper reporting one fixed `split_weight` everywhere and asserts deterministic pair-count bounds (max task share ≤ 10 %; 23–25 % before, 5–6 % after).

**Before/after** (shared 64-core box, t8, back-to-back same session, loadavg ~29–36; GLO-30 → IGEO7 covering; "main" = v0.2.9 89ec5ca):

| eager plan | wall | util | max task share | corr(est, actual) |
|:--|--:|--:|--:|--:|
| 512² frontier@508a637 | 6.54 s | 2.78 | 50.0 % | 0.20 |
| 512² **fixed** | **2.63 s** | **6.78** | **1.6 %** | 0.42* |
| 512² CR main | 2.67 s | 6.98 | 0.1 % | – |
| 1024² frontier@508a637 | 22.83 s | 2.84 | 50.0 % | 0.21 |
| 1024² **fixed** | **9.24 s** | **6.78** | **1.6 %** | 0.36* |
| 1024² CR main | 9.74 s | 6.78 | 0.4 % | – |

\* with DGG's exact `split_weight` methods (JuliaGeo/DiscreteGlobalGrids.jl PR); with the fallback alone corr is 0.20/0.28 but the share and wall figures are identical — the conservation rule doesn't need honest weights to balance.

**No regression** where the frontier already performed: DGG-native source plans within noise of 508a637 back-to-back (d512 19.6 s → 20.0 s, d1024 75.8 s → 75.6 s, util 6.9–7.2, identical digests), and the lat-lon/HEALPix/cubed-sphere suite paths are covered by the tests.

**Identity**: pair lists byte-identical to serial `dual_depth_first_search` at `chunks_per_thread ∈ {1,2,8,32}` and weight matrices bit-identical (`colptr`/`rowval`/`nzval`) on both RasterGrid→DGGS and DGGS→DGGS at t8; end-to-end regrid digests unchanged across arms.

**Suite**: 10053 pass / 0 fail at `JULIA_NUM_THREADS=8` — the 10049 baseline plus exactly the 4 new assertions.

https://claude.ai/code/session_01SwnXToXzkTdstjqfp8ovhH

PR #138, opened 2026-08-20
Parallel sparse assembly: build the CSC from column windows

asinghvi17:
Stacks on #137 (`claude/budget-frontier`); base is that branch, not `main`.

## What was serial

`intersection_areas` assembled the COO triplets on `npartitions` tasks and then ran two
strictly serial passes over the whole matrix: `reduce(vcat, ...)` merging the per-task
vectors, then one `sparse()` over the merged coordinate list. On a 66,129,312-entry
regrid (a 3600² Copernicus DEM tile onto 16,181,892 IGEO7 cells) that tail is **2.4-2.7 s
on one thread**, and it does not shrink with more threads.

## What is parallel now

A coarse column histogram (4096 blocks, counted per chunk in parallel) cuts the columns
into entry-balanced windows. One threaded pass buckets the triplets straight into their
window — no concatenation — and each window is built by `SparseArrays.sparse!` on its own
task, then spliced into the final `colptr`/`rowval`/`nzval`.

`sparse`'s output for a column is a function of that column's entries and their input
order alone: it counting-sorts to CSR, folds repeats within a row in input order, and
counting-sorts back. A window keeps a column's entries and their global order, so its
columns come out identical to the ones the single `sparse()` call produced.

Windows are capped at one per thread and at `nnz / nrows`, which keeps the windows'
`O(nrows)` counting-sort scratch inside the size of the entries they sort; below two
windows the old concatenate-and-`sparse` route runs unchanged. Each window owns its COO
arrays outright, so `sparse!` takes the CSC's storage from them (the documented
`sparse!(I, J, V, …, I, J, V)` reuse) instead of allocating it a second time.

`assemble_sparse_matrix_coo` keeps its concatenating behaviour for outside callers.

## Bit-identity

Not "equal" — `colptr`, `rowval` and the `nzval` bit patterns.

- New tests: 6 randomised shapes (200×5000 … 300×100000, 7-32 chunks) each with a third
  of every chunk repeated so whole columns carry duplicate coordinates within and across
  chunks, swept over 2/3/8 windows; plus a skewed column distribution and chunks that
  produced nothing. `nnz(expected) < length(rows)` asserts the fixtures really do dedup.
- The real matrix: on the 32 captured per-task triplets of the regrid above,
  `colptr ==`, `rowval ==`, `reinterpret(UInt64, nzval) ==` at 2, 4 and 8 windows,
  nnz 66,129,312.
- End to end: the DGG eager MOC regrid digest `sum = 3.368822268175383e10`,
  `sha = 5b696475a3665634`, `n = 16181892` on all 16 reps of the A/B below, both arms.

Existing suite: 10200/10200 pass, Julia 1.12, `-t 8`.

## Numbers

Shared 64-core box, `/proc/loadavg` beside every measurement, both arms in one window.

The tail alone, measured in situ (real trees, real chunks, real heap, the two tails raced
against each other in one process, 3 interleaved reps, load 92-104):

| tail | reps (s) | churn | GC |
|:--|:--|--:|--:|
| `reduce(vcat)` + `sparse` | 2.666 / 2.674 / 2.453 | 3.76 GiB | 0.004-0.007 s |
| column windows | **1.263 / 1.179 / 1.089** | 4.13 GiB | 0.041-0.169 s |

**−1.42 s (−55 %).**

Whole warm regrid, `-t 8`, four ABBA passes of 1 cold + 3 warm reps, load 46-78:

| arm | warm reps (s) | churn | maxrss |
|:--|:--|--:|--:|
| `claude/budget-frontier` @ 6a4b997 | 51.13 / 49.79 / 49.28 · 46.80 / 49.13 / 46.99 | 30.58 GiB | 9.5-10.3 GiB |
| this branch | 49.27 / 48.34 / 48.61 · 46.28 / 47.38 / 45.58 | 30.95 GiB | 7.4-8.2 GiB |

Mean 48.85 → 47.58 s, **−1.27 s (−2.6 %)**, churn +0.37 GiB, peak RSS −2.0 GiB (the serial
route holds the concatenated COO and `sparse`'s scratch at the same time). The box carried
load 46-78 throughout, so the whole-call spread is ±2 s; the in-situ pairing above is the
measurement that resolves the lever, and the two agree.

The tail is invariant in thread count, so it is worth proportionally more at higher counts
— on this workload it is ~15 % of a t8 call and ~37 % of a t64 one.

https://claude.ai/code/session_01SwnXToXzkTdstjqfp8ovhH


asinghvi17:
## Codex CLI review (gpt-5.6-sol, high reasoning) — diff vs `origin/claude/budget-frontier`

Ran from a fresh clone at `0315553`, with the test env instantiated (Julia 1.12.6, SparseArrays 1.12.0). Codex was told not to read only — it built its own adversarial fixtures and compared against `sparse` over the concatenation bitwise via `reinterpret(UInt64, nzval)`.

**No valid-input bit-identity or threading defect was found.** Clean: all entries in one column; one entry per column; 200×500 000 with 20 000 entries; a 200 000×17 tall matrix; duplicates spanning chunks with `1e16, 1.0, -1e16`; subnormals and `±0.0`; window cuts inside a hot duplicate run; empty / interspersed / all-empty chunks; one row; one column; fewer entries than threads; `Bool` duplicates under `|`; 500 seeded randomised differential cases; key fixtures repeated 100× at `-t 8` with no nondeterminism. It also read the pinned SparseArrays source and confirmed the `sparse!(…, I, J, V)` output-aliasing pattern in `_window_csc` is explicitly documented, with every scratch-length precondition satisfied and every COO read finished before the aliased outputs are resized.

Five findings, and my disposition:

| # | Sev | Finding | Disposition |
|---|-----|---------|-------------|
| 1 | major | `_bucket_chunk!` walks `cols` under `@inbounds` while indexing `rows[j]`/`vals[j]`; nothing checks that a chunk's three vectors have equal length. `sparse` throws `ArgumentError` on ragged triplets; the windowed route silently truncated (`[([1,1],[1],[2.0,3.0])]` → `nzval == [2.0]`) or read out of bounds. Reachable from the public `InPlace` extension point, where user code pushes onto the three vectors itself. | **REAL — fixed** in 3b50a2f: `_check_chunk_lengths` runs once per chunk at the top of `_sparse_from_chunks`, so both the windowed and the serial-fallback route reject it, with a regression test. |
| 2 | minor | A zero-length *vector of chunks* hits `chunks[1][3]` and throws `BoundsError`. | **WON'T FIX (by design).** The value type is not recoverable from an empty chunk list — that is exactly why the guard lives one level up in `_assemble_sparse`, which has `output_eltype(op, src_tree, dst_tree)`. Pushing it down would mean threading `Tv` through a private helper for no caller. |
| 3 | minor | Forcing `nwindows ≥ 2` with `nrows == 0` reaches the expert `sparse!`, which reads `csrrowptr[2]` out of a length-`m+1 == 1` array. | **WON'T FIX — unreachable.** The default selector is `min(nthreads(), ntotal ÷ max(nrows, 65536))`; with `nrows == 0` that needs ≥ 131 072 entries, every one of which would carry a row index outside `1:0` and be rejected by the serial route too. Only the explicit `nwindows` overload used by the tests can get there. |
| 4 | nit | The comment on `_count_blocks!` claimed it is where an out-of-range column is caught; with `blockwidth > 1`, `col == 0` maps into block 1 and columns just past `ncols` fall in the last block's padding. | **REAL (doc) — fixed.** Comment now says what the bounds check is actually for: `_bucket_chunk!` indexes `windowof` with these block numbers under `@inbounds`. Such columns still cannot corrupt anything — they land in a window whose local index `sparse!` then rejects. |
| 5 | minor | The committed tests are random fixtures with forced window counts; they do not directly encode cross-chunk cancellation, signed zeros/subnormals, `Bool` combine, or the default selector. | **REAL — fixed.** New `Windowed COO assembly: directed cases` testset: three duplicates of one coordinate split across three chunks summing to `0.0` in input order and `1.0` in any other, `-0.0 + 0.0`, subnormals, `Bool`'s `|`, the ragged chunk, and the 3-arg default. |

### Gates after the fix

- **CR suite**, Julia 1.12, `-t 8`: **10233 / 10234**. The one failure is `Extensions: ClimaCore → se_field_to_vec and vec_to_se_field!` (`isapprox(0.0, -0.75)` — `sum(field)`'s reduction order inside ClimaCore). It is **pre-existing**: re-running the suite with `src/regridder/intersection_areas.jl` restored from `origin/claude/budget-frontier` reproduces it identically, and the test touches no assembly code at all. It also passes 3/3 when that file is run standalone.
- **DGG digest gate** (level-13 IGEO7 MOC covering, 3600² Copernicus GLO-30, eager conservative regrid), DGG `1e41af2` with this checkout `Pkg.develop`ed in: `n = 16181892`, `sum = 3.368822268175383e10`, `sha = 5b696475a3665634` — **exact match**, 60.0 s at `/proc/loadavg` 54.8 → 60.1 on the shared 64-core box.

New head: `3b50a2f`.

src/regridder/intersection_areas.jl
```diff
@@ -196,21 +196,222 @@ function assemble_sparse_matrix_coo(style::S, op::O, items::I, src_tree::T1, dst
...
-    vals = reduce(vcat, getindex.(all_results, 3))
-    return rows, cols, vals
+    return map(fetch, result_tasks)
+end
+
+# Concatenate the per-chunk COO vectors into single vectors, in partition order.
+# `init` would cost `vcat`'s pre-sized specialisation, making this quadratic in the partition count.
+_concat_chunks(chunks) = (reduce(vcat, getindex.(chunks, 1)),
+                          reduce(vcat, getindex.(chunks, 2)),
+                          reduce(vcat, getindex.(chunks, 3)))
```
asinghvi17: This does not need to exist.

PR #140, opened 2026-08-22
Replace cached dual DFS with weighted leaf carry

asinghvi17:
Stacked on #138 (which stacks on #137). This PR contains only work above `claude/parallel-assembly`.

## Summary

- Replaces `CachedDualDepthFirstSearch` with deterministic, one-sided `WeightedDualDepthFirstSearch`.
- The traversal always opens the heavier non-leaf side (first side on ties) and carries eager leaf `child_indices_extents` through the opposite descent using lazy `@something`; no retained extent cache or scratch stack remains.
- Serial candidate collection and the threaded frontier use the same split rule. Threaded flattening sizes the result exactly before copying.
- Adds bounded, eager `SmallCollections` leaf entries for `QuadtreeCursor`, `TopDownQuadtreeCursor`, `IndexOffsetQuadtreeCursor`, `ReorderedTopDownQuadtreeCursor`, and native HEALPix. Reordered cursor fields are now type-stable.
- Updates traversal/cursor/extension tests and adds `SmallCollections` dependencies.

## Correctness and performance

On the production DGG benchmark (8 threads, exact 3600×3600 COPDEM tile → 16,172,725 IGeo7 L13 cells), candidate traversal emits exactly 96,863,960 candidates. Its multiset fingerprint matches the cached baseline:

- sum: `aeb070ff639111d8`
- xor: `2ab7e7468073fc8e`
- sumsq: `77e48d9051c323d2`

Warm candidate traversal/task-local creation is 6.06 s plus 0.75 s flattening, versus approximately 9.9 s for the old cached traversal/task-local work. Full regrid is 20.54 s (second run 22.24 s), back near the original 20 s.

## Validation

- Focused weighted traversal tests: 32/32 under both ConservativeRegridding and DGG dependency sets.
- Native HEALPix extension tests: 34/34.
- Core eager leaf objects are inferred, `isbits`, and allocate 0 bytes warm.
- This checkout reran the focused weighted test (32/32) and native HEALPix test (34/34).

## Known dependency caveat

Do not treat the standalone ConservativeRegridding broad whole-globe suite as green: its stale released GeometryOps 0.1.44 lacks `Extents.intersects(::SphericalCap, ::SphericalCap)`. Production DGG pins newer GeometryOps; the Float32 global weighted self-search returns 320 candidates there.


test/extensions/climacore.jl
```diff
@@ -50,7 +50,12 @@ const ClimaCoreExt = Base.get_extension(ConservativeRegridding, :ConservativeReg
 
         # Weighted integral via sum(field) should match manual dot(weights, values)
         weights = ClimaCoreExt.se_node_weights(cubedsphere_space)
-        @test isapprox(sum(weights .* v), sum(field), rtol=1e-10)
+        weighted_values = weights .* v
+        # The latitude integral is theoretically zero, so a relative-only tolerance is
+        # ill-posed: different valid reduction orders leave O(1) residuals after
+        # cancelling O(10^16) absolute contributions on an Earth-radius sphere.
+        cancellation_atol = 10 * eps(eltype(weighted_values)) * sum(abs, weighted_values)
```
asinghvi17: Should this not be eps(sum(abs, weighted_values))?  To more accurately reflect the actual problem.

For reference
```julia
julia> eps()
2.220446049250313e-16

julia> eps(1e16)
2.0
```

test/extensions/oceananigans.jl
```diff
@@ -3,7 +3,25 @@ using ConservativeRegridding: Trees
...
+    padding_polygon = Trees.getcell(grid, 1, 1)
+    padded = OceananigansExt.PaddedTreeWrapper(real, 2, padding_polygon, 10)
+
+    # The tree emits only the real IDs 11:12, but its addressable output domain
+    # also contains padded IDs 13:14, which remain zero in sparse assembly.
+    @test vec(first.(collect(STI.child_indices_extents(padded)))) == [11, 12]
+    @test Trees.ncells(padded) == (4, 1)
+    @test Trees.cell_index_count(padded) == 14
+    @test Trees.getcell(padded, 13) === padding_polygon
+    @test Trees.getcell(padded, 14) === padding_polygon
```
asinghvi17: This should probably get all leaf nodes via a simple walk down the tree, then test those?

test/trees/grids.jl
```diff
@@ -203,34 +203,261 @@ import GeometryOps as GO
...
+    ir, jr = 2:3, 1:2
+    imin, imax = first(ir), last(ir) + 1
+    jmin, jmax = first(jr), last(jr) + 1
+    expected_center = LinearAlgebra.normalize(
+        (_SPH_PTS[imin, jmin] + _SPH_PTS[imax, jmin] +
+         _SPH_PTS[imax, jmax] + _SPH_PTS[imin, jmax]) / 4)
+
+    PP = ConservativeRegridding.Trees.CurvilinearGridPerimeterPoints
+    raw_radius = maximum(p -> GO.spherical_distance(expected_center, p),
+        PP(_cellbased_sph(), imin, imax, jmin, jmax))
```
asinghvi17: Construct the grid on the line above.  I don't like this PP thing either.

asinghvi17: Construct the grid on the line above.  I don't like this PP thing either.

test/cached_dual_dfs.jl
```diff
@@ -0,0 +1,294 @@
+using Test
+using ConservativeRegridding
+using ConservativeRegridding.Trees
+import ConservativeRegridding as CR
+import GeoInterface as GI, GeometryOps as GO
+import GeometryOps: SpatialTreeInterface as STI
+import Extents
+
+const CDFS = ConservativeRegridding.CachedDualDepthFirstSearch
```
asinghvi17: use import instead.

bench/quadtree_cursor.jl
```diff
@@ -0,0 +1,61 @@
...
+# JULIA_LOAD_PATH. The default 896×896 half-cell-shifted grids make the stock
+# cursor's child construction visible while keeping a five-sample run short.
+
+import ConservativeRegridding as CR
+import Extents
+import GeometryOps as GO
+
+const N = parse(Int, get(ENV, "CR_CURSOR_BENCH_N", "896"))
+const LEAF = parse(Int, get(ENV, "CR_CURSOR_BENCH_LEAF", "4"))
+const REPS = parse(Int, get(ENV, "CR_CURSOR_BENCH_REPS", "5"))
```
asinghvi17: No.  Make these local things to the script.  No env hacks ideally.

bench/quadtree_cursor.jl
```diff
@@ -0,0 +1,61 @@
...
+tree1, tree2 = if hasproperty(prototype1, :leafsize)
+    (CR.Trees.TopDownQuadtreeCursor(grid1; leafsize = (LEAF, LEAF)),
+     CR.Trees.TopDownQuadtreeCursor(grid2; leafsize = (LEAF, LEAF)))
+else
+    (prototype1, prototype2)
+end
+
+mutable struct Counter
+    n::Int
+end
```
asinghvi17, line 15: a little description would be helpful

bench/quadtree_cursor.jl
```diff
@@ -0,0 +1,61 @@
+# Microbenchmark for the raster-shaped TopDownQuadtreeCursor hot path.
+#
+# Run this file with the ConservativeRegridding checkout under test first in
+# JULIA_LOAD_PATH. The default 896×896 half-cell-shifted grids make the stock
+# cursor's child construction visible while keeping a five-sample run short.
```
asinghvi17: Run under the docs env, which already has the right linkages.   Any packages you need (and please do benchmark under Chairmarks.jl) should be added there.

bench/quadtree_cursor.jl
```diff
@@ -0,0 +1,61 @@
...
+step = 1 / N
+x1 = collect(range(0.0, 1.0; length = N + 1))
+x2 = x1 .+ step / 2
+y = collect(range(0.0, 1.0; length = N + 1))
+grid1 = CR.Trees.RegularGrid(GO.Planar(), x1, y)
+grid2 = CR.Trees.RegularGrid(GO.Planar(), x2, y)
+
+prototype1 = CR.Trees.TopDownQuadtreeCursor(grid1)
+prototype2 = CR.Trees.TopDownQuadtreeCursor(grid2)
+tree1, tree2 = if hasproperty(prototype1, :leafsize)
```
asinghvi17: What do you mean _if_?

src/regridder/intersection_areas.jl
```diff
@@ -122,20 +125,72 @@ should_store_result(result) = error("""
...
+
+# Repeated serial builds in an outer worker keep these buffers on that task.  A task
+# may migrate between scheduler threads, so thread-indexed storage would be unsafe.
+mutable struct _AssemblyScratch{T}
+    candidate_pairs::Vector{Tuple{Int, Int}}
+    rows::Vector{Int}
+    cols::Vector{Int}
+    vals::Vector{T}
+    in_use::Bool
+end
```
asinghvi17: Let's call this SparseMatrixAssemblyCache and make its existence public api, but not the fields necessarily.  The only public thing should be, that you can pass this through into whichever function uses it as `cache = ...`.

src/regridder/intersection_areas.jl
```diff
@@ -122,20 +125,72 @@ should_store_result(result) = error("""
...
+
+function _release_assembly_scratch!(scratch::_AssemblyScratch)
+    empty!(scratch.candidate_pairs)
+    empty!(scratch.rows)
+    empty!(scratch.cols)
+    empty!(scratch.vals)
+    scratch.in_use = false
+    return nothing
+end
```
asinghvi17: this should be `empty!(scratch)`

src/regridder/intersection_areas.jl
```diff
@@ -122,20 +125,72 @@ should_store_result(result) = error("""
...
+
+_AssemblyScratch(::Type{T}) where {T} =
+    _AssemblyScratch(Tuple{Int, Int}[], Int[], Int[], T[], false)
+
+# The key is a module-private type object, unique for every COO value type without
+# allocating a composite key on each lookup.
+struct _AssemblyScratchKey{T} end
+
+function _acquire_assembly_scratch(::Type{T}) where {T}
+    tls = task_local_storage()
```
asinghvi17: that only works on julia 1.12.  use TaskLocalValues.jl instead I guess.

asinghvi17: and this looks like it suffers from TOCTTOU also.  maybe the scratch should be managed by the regridder function or the chunkwindows or something which can instantiate it and pass it on.  but it should be actively managed.

src/trees/grids.jl
```diff
@@ -244,37 +257,45 @@ arbitrary 4-corner polygons. Internally a thin wrapper over [`_spherical_cap`].
...
     return _spherical_cap(p1, p2, p3, p4,
         (GO.UnitSphereFromGeographic()(p) for p in other_points))
 end
 
-# Build a SphericalCap covering 4 CCW corners (SW, SE, NE, NW), the slerp
-# midpoints of their 4 edges (great-circle bulge), and every point yielded by
-# `perimeter` (constant-lat/lon bulge along cell sides).
+# Build a SphericalCap covering 4 CCW corners (SW, SE, NE, NW) and every point
+# yielded by `perimeter`. Caps wider than a hemisphere are not geodesically
+# convex, so only narrower caps can conservatively stand in for their edges.
```
asinghvi17: > Caps wider than a hemisphere are not geodesically convex

Yes but no one said that the grid had to be convex - only the individual cells.  It's perfectly possible for me to make a concave grid with convex cells.

src/trees/interfaces.jl
```diff
@@ -255,6 +274,24 @@ STI.isleaf(cursor::AbstractQuadtreeCursor) = error("GO.STI.isleaf not implemente
...
+rather than reading it off a couple of coordinate vectors.  True for every grid but the
+planar regular one, where the answer is four array reads.
+
+This is the grid-level fact behind `STI.node_extent_is_expensive` for the quadtree
+cursors: a cursor's `node_extent` is exactly `cell_range_extent` over its leaf range, so
+it is expensive precisely when the grid's is.  The search in
+[`ConservativeRegridding.CachedDualDepthFirstSearch`](@ref) reads the cursor-level trait
+to decide whether caching a node's child extents pays for itself.
+
+Defined on the type, so it stays a compile-time constant; the instance method forwards.
```
asinghvi17: DECREASE TEXT LENGTH this is a wall of text

src/utils/MultithreadedDualDepthFirstSearch.jl
```diff
@@ -204,34 +207,47 @@ function _spawn_pair(inner::IF, predicate::P, node1::N1, node2::N2) where {IF, P
...
     tasks = map(pairs) do pair
         _spawn_pair(_inner_dfs_f, predicate, pair[1], pair[3])
     end
-    results = Vector{Tuple{Int, Int}}[fetch(t)::Vector{Tuple{Int, Int}} for t in tasks]
-    # `init` would cost `vcat`'s pre-sized specialisation, making this quadratic in the task count.
-    return isempty(results) ? Tuple{Int, Int}[] : reduce(vcat, results)
+    # Fetch and append in frontier order, preserving the serial DFS order.  The caller
+    # may retain `result` in task-local scratch across repeated block builds.
+    for task in tasks
+        append!(result, fetch(task)::Vector{Tuple{Int, Int}})
```
asinghvi17: Can drop the :: I think here

asinghvi17 (commented):
Please address these comments.

